PR #10, opened 2022-02-16
Add a python node to publish generated random images

Purg:
Somewhat of a testbed for benchmarking image transmission.


ros/README.md
```diff
@@ -0,0 +1,11 @@
+ROS2 workspace root.
+
+# Build workspace
+```bash
+$ rosdep install -i --from-path . --rosdistro foxy -y
+$ colcon build
```
joshanderson-kw, line 6: In your environment, after you do `colcon build`, do you also have to source the `install/local_setup.bash` file before you do `ros2 run...`?

Purg: I do, but in a different terminal given advice that I read to not mix the build-env with the install-sourced-env. I seem to not have gotten to writing that, which I can

joshanderson-kw (approved):

ros/README.md
Purg: I added a second little bit of info here. Its not a good solution, but reflective of what I did. Your incoming docker stuff likely addresses some of this.

joshanderson-kw (approved):

PR #11, opened 2022-02-16
Dev/add object detector docker node

joshanderson-kw:


docker/object_detector/Dockerfile
```diff
@@ -0,0 +1,71 @@
...
+    LANG=C.UTF-8
+
+# Add ``object_detector`` user
+RUN useradd -mr -s /bin/bash object_detector \
+ # sudo permission for modifying permissions at runtime (see entrypoint.sh).
+ # TODO: Remove this eventually since its defeating the point of having a user.
+ && echo "object_detector ALL=(ALL:ALL) NOPASSWD:ALL" >>/etc/sudoers
+
+USER object_detector
+WORKDIR /home/object_detector
```
Purg, line 52: Adding an in-container user at this stage might just development more complicated unless you have an established motivation.

docker/object_detector/Dockerfile
```diff
@@ -0,0 +1,71 @@
+##############################################################################
+# Base system packages and depedencies
+#
```
Purg, line 3: I think we want to start with one general development node. We're still early to focus down into creating node-specific containers. We can use a general base note to run specific nodes in our built workspace.

docker/object_detector/Dockerfile
```diff
@@ -0,0 +1,71 @@
...
+FROM nvidia/cuda:9.2-cudnn7-runtime-ubuntu18.04 AS ptg-gpu
+CMD nvidia-smi
```
Purg, line 5: This intermediate build doesn't seem to be used or referenced elsewhere? Am I missing it?

ros/cyclonedds_profile.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<CycloneDDS xmlns="https://cdds.io/config" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="https://cdds.io/config https://raw.githubusercontent.com/eclipse-cyclonedds/cyclonedds/master/etc/cyclonedds.xsd">
+    <Domain Id="any">
+        <General>
+            <NetworkInterfaceAddress>192.167.1.90</NetworkInterfaceAddress>
```
Purg, line 5: This is too specific to check in and is only local to your machine/network.

unity/Hello_World/ProjectSettings/ProjectSettings.asset
```diff
@@ -193,6 +193,7 @@ PlayerSettings:
   - {fileID: 0}
   - {fileID: 0}
   - {fileID: 0}
+  - {fileID: 0}
```
Purg, line 196: This doesn't seem related to anything? I don't know why changes like this happen or what they mean.

PR #12, opened 2022-02-16
Add linux support to cpp publisher

joshanderson-kw:
Tested on Ubuntu and Windows PC

### DIFF
--- a/ros/cpp/src/cpp_pubsub/src/publisher_member_function.cpp
+++ b/ros/cpp/src/cpp_pubsub/src/publisher_member_function.cpp
@@ -15,6 +15,7 @@
 
 #include "sensor_msgs/msg/image.hpp"
 
+#ifdef _WIN32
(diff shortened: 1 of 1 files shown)


ros/cpp/src/cpp_pubsub/src/publisher_member_function.cpp
```diff
@@ -87,8 +100,8 @@ class MinimalPublisher : public rclcpp::Node
 
   private:
     size_t count_;
-    std::string server_ip_addr = "169.254.103.120";
-    
+    std::string server_ip_addr = "192.167.1.90";
```
Purg, line 103: OK for now but we need to make this a node parameter in the future. https://docs.ros.org/en/foxy/Tutorials/Using-Parameters-In-A-Class-CPP.html#

Purg (approved):

PR #13, opened 2022-02-17
Dev/swap tcp servers and clients

joshanderson-kw:


joshanderson-kw:
TCP servers now started on headset and ROS publisher now contains the TCP clients. 
IP address is still hard-coded in the ROS publisher, but the headset uses the `NetworkInterface` class to get the IPv4 address of the Hololens at runtime. It defaults to the network interface that is `Ethernet` and its status is `Connected`. Hopefully, this means we don't have to make special HL builds for new IP addresses.

PR #16, opened 2022-02-24
Add app publish and device portal install instructions

joshanderson-kw:


unity/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+3) In the app package window, select `Sideloading`, uncheck the `Enable automatic updates` box, and click `Next`.
+
+![Distribution screen](docs/images/distribution_method.PNG)
+
+4) Ensure the app certificate is signed and click `Next`. The default certificate settings are OK.
+
+![Signing screen](docs/images/signing_method.PNG)
+
+5) Choose the app's output location and ensure the ARM64 architecture is selected. Click `Create` to create the package.
```
Purg, line 15: Does deselecting x86, x64, 32-bit ARM decrease package size / installation time? I am assuming we can just target the one x64 ARM because that's all the HL2 is.

Purg (approved):
One minor question/comment, otherwise LGTM

unity/README.md
joshanderson-kw: Yeah, it's quite a bit faster to only build ARM64. I added a note.

PR #18, opened 2022-03-02
Add object deteciton set message, update py-nodes, launch file demo

Purg:
See `ros/angel_debug/README.md` for running the launch file demo. Will want to connect to your video device ID 0, likely your webcam.


joshanderson-kw (approved):
Worked for me!

PR #21, opened 2022-03-02
Dev/fetch headset spatial data

joshanderson-kw:
Adds spatial mapping mesh streaming and includes the world/projection matrices with each RGB image frame.


joshanderson-kw:
Adds spatial mapping mesh streaming and includes the world/projection matrices with each RGB image frame.

Purg (commented):
At some point we will likely meed to do something about the quantity of binary files stored in the unity space: e.g. the added `com.microsoft.mixedreality.sceneunderstanding-0.6.0.tgz` file.

unity/Hello_World/Assets/MixedRealityToolkit.Generated/CustomProfiles/New MixedRealitySpatialAwarenessMeshObserverProfile.asset
```diff
@@ -0,0 +1,29 @@
+%YAML 1.1
```
Purg, line 1: Why are there 3 versions of this file? Are they all needed?

unity/Hello_World/Assets/Scripts/SpatialMappingCapture.cs
```diff
@@ -0,0 +1,250 @@
...
+using System.Net.NetworkInformation;
+using System.Net.Sockets;
+using System.Threading;
+using System.Threading.Tasks;
+using UnityEngine;
+
+
+//using SpatialAwarenessHandler = IMixedRealitySpatialAwarenessObservationHandler<SpatialAwarenessMeshObject>;
+
+//public class SpatialMappingCapture : DemoSpatialMeshHandler, IMixedRealitySpatialAwarenessObservationHandler<SpatialAwarenessSceneObject>
```
Purg, line 21: Lots of commented out code throughout. Can we remove most of this? If we keep any around, we should add comments as to why.

unity/Hello_World/Assets/Scripts/SpatialMappingCapture.cs
```diff
@@ -0,0 +1,250 @@
...
+
+        tcpServer = new TcpListener(localAddr, SMTcpPort);
+
+        // Start listening for client requests.
+        tcpServer.Start();
+
+        // Perform a blocking call to accept requests.
+        // You could also use server.AcceptSocket() here.
+        tcpClient = tcpServer.AcceptTcpClient();
+        tcpStream = tcpClient.GetStream();
```
Purg, line 106: Not really a comment about this PR, however could this function, which seems to be forked off to a thread to process, also continue on heart-beat or something to check for new client connections? I'm thinking of being robust to the ROS node receiver crashing / shutting-down / restarting use-case.

unity/Hello_World/Packages/manifest.json.backup3
```diff
@@ -0,0 +1,59 @@
+{
```
Purg, line 1: what are these "backup3" files? Do they really need to be committed or are they things that should be ignored?

unity/Hello_World/Assets/MixedRealityToolkit.Generated/CustomProfiles/New MixedRealitySpatialAwarenessMeshObserverProfile.asset
joshanderson-kw: No... was experimenting with different mesh observer profiles. I removed the unused ones.

unity/Hello_World/Assets/Scripts/PVCameraCapture.cs
```diff
@@ -109,18 +122,68 @@ async void Start()
...
         int targetVideoWidth, targetVideoHeight;
         float targetVideoFrameRate;
-        targetVideoWidth = 1280;
-        targetVideoHeight = 720;
-        //targetVideoWidth = 1920;
-        //targetVideoHeight = 1080;
+        //targetVideoWidth = 1280;
+        //targetVideoHeight = 720;
+        targetVideoWidth = 1920;
+        targetVideoHeight = 1080;
```
Purg, line 184: Is this an arbitrary increase in resolution, or was this required for the "fix" to world localization? I'm not opposed, just curious / trying to understand things better.

unity/Hello_World/Assets/Scripts/SpatialMappingCapture.cs
joshanderson-kw: I think that could work. I made it its own thread because if it's on the main unity thread, the `AcceptTCPClient()` call will block the thread until the TCP connection is established, which means nothing is rendered to the headset and it looks like nothing is happening.

unity/Hello_World/Assets/Scripts/PVCameraCapture.cs
joshanderson-kw: It was one of the things I tried to fix the localization issue. It didn't end up fixing the issue, but it didn't hurt so I left it in.

unity/Hello_World/Assets/Scripts/SpatialMappingCapture.cs
Purg: > I made it its own thread ...

No argument there, more thinking that this separate thread could continue existing and loop, re-attempting tcp client acceptance after the current one "closes", assuming that is appropriately detected.

Of course, it's likely the case that the potential effort doing that should be expended moving to something like 0mq/redis/etc. that will handle the communications (including reconnections, etc.) for us.

Purg (approved):

PR #22, opened 2022-03-02
Dev/add spatial map node

joshanderson-kw:
Note: this PR depends on the unity changes from PR #21 , so it is marked as draft for now.

Add spatial map ROS2 node.
Add libgl docker install so we can use trimesh plotting.
Started cleaning up the talker node to get away from byte arrays by creating two new angel_msgs: SpatialMesh and HeadsetPose.
Rename cpp_pubsub to ptg_datahub.
Convert NV12 image to RGB8 image in talker node.


docker/workspace-base-dev/Dockerfile
```diff
@@ -24,3 +25,18 @@ RUN apt-get -y update \
...
+    libgl1 \
+    libglx0 \
+    libegl1 \
+    libxext6 \
+    libx11-6 \
+  && rm -rf /var/lib/apt/lists/*
+
+# Env vars for the nvidia-container-runtime.
+ENV NVIDIA_VISIBLE_DEVICES all
+ENV NVIDIA_DRIVER_CAPABILITIES graphics,utility,compute
```
Purg, line 42: I'm curious to learn what these are for. This is possibly the correct location (though merged with the apt-install earlier in this file). Is this required for the trimesh python package stack?

ros/cpp/src/ptg_datahub/CMakeLists.txt
```diff
@@ -15,18 +15,26 @@ find_package(rclcpp REQUIRED)
...
-
-
+find_package(geometry_msgs REQUIRED)
+find_package(shape_msgs REQUIRED)
+find_package(angel_msgs REQUIRED)
+find_package(cv_bridge REQUIRED )
+find_package(OpenCV REQUIRED COMPONENTS core )
+
+add_executable(talker src/ptg_datahub.cpp)
+ament_target_dependencies(talker
```
Purg, line 25: This node could likely be renamed to "datahub" as well instead of the "hello world" "talker" example name. I mostly just don't think "talker" is appropriate anymore.

ros/cpp/src/ptg_datahub/src/ptg_datahub.cpp
```diff
@@ -81,7 +89,7 @@ class MinimalPublisher : public rclcpp::Node
     std::thread StartTCPServerThread(int port)
```
Purg, line 89: Maybe not a "this branch" thing, but I think something better could be used here as an identifier than the port integer. E.g. we could make an enum of the different sensor stream "types" and take in a value from that enum here. Then we have some center to index into various mappings (the enum values) and have a natural constraint in the values that can be provided to server threads: enum values (e.g. don't hvae to change this port-range check when expanding/changing sensor port attribution).

ros/cpp/src/ptg_datahub/src/ptg_datahub.cpp
```diff
@@ -113,6 +126,11 @@ class MinimalPublisher : public rclcpp::Node
       unsigned int frames_received = 0;
 
       publisher_ = this->create_publisher<sensor_msgs::msg::Image>(PORT_TOPIC_MAP[port], 10);
+
+      if (port == PV_TCP_PORT)
+      {
+        pose_publisher_ = this->create_publisher<angel_msgs::msg::HeadsetPoseData>("HeadsetPoseData", 10);
```
Purg, line 132: Red flag: Defining a constant in the body of the code. Instead create a constant up at the top with everything else. Eventually when all this stuff becomes variables, then we don't have to hunt for things like this.

ros/cpp/src/ptg_datahub/src/ptg_datahub.cpp
```diff
@@ -311,6 +355,160 @@ class MinimalPublisher : public rclcpp::Node
...
+            memcpy(&t.vertex_indices[2], &frame_data[triangle_offset + i + 8], sizeof(t.vertex_indices[2]));
+            triangle_list.insert(triangle_list.end(), t);
+          }
+
+          message.mesh.vertices = vertex_list;
+          message.mesh.triangles = triangle_list;
+        }
+
+        publisher_->publish(message);
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
Purg, line 508: Maybe a silly question that I didn't think to ask before: why do these `TCPServer*Thread` loops have millisecond sleeps in them?

ros/cpp/src/ptg_datahub/src/ptg_datahub.cpp
```diff
@@ -384,6 +581,7 @@ int main(int argc, char * argv[])
   t8.join();
   t9.join();
   t10.join();
+  t11.join();
```
Purg, line 584: This is fine for now, but it might be better to start/join these threads as part of the node-class operation, e.g. start in constructor, join in destructor.

ros/python/py_pubsub/py_pubsub/object_detector.py
```diff
@@ -56,6 +57,8 @@ def __init__(self):
         self._detector = ResNetFRCNN(img_batch_size=1, use_cuda=self._use_cuda)
         log.info(f"Ready to detect using {self._detector}")
 
+        self.stop = False
```
Purg, line 60: What is this variable for? Nothing seems to ever set it to anything than this default value. I presumed that the detector node "stop" signal will come in the form of terminating/killing the containing process. Do you have a different use-case in mind?

ros/python/py_pubsub/py_pubsub/spatial_mapper.py
```diff
@@ -0,0 +1,311 @@
...
+    def __init__(self):
+        super().__init__(self.__class__.__name__)
+
+        self.declare_parameter("spatial_map_topic", "SpatialMapData")
+        self.declare_parameter("det_topic", "ObjectDetections")
+        self.declare_parameter("pose_topic", "HeadsetPoseData")
+
+        self._spatial_map_topic = self.get_parameter("spatial_map_topic").get_parameter_value().string_value
+        self._det_topic = self.get_parameter("det_topic").get_parameter_value().string_value
+        self._pose_topic = self.get_parameter("pose_topic").get_parameter_value().string_value
```
Purg, line 30: I learned that the `declare_parameter` method also returns the parameter instance, so we can collapse these two bits into one.

```suggestion
        self._spatial_map_topic = self.declare_parameter("spatial_map_topic", "SpatialMapData").get_parameter_value().string_value
        self._det_topic = self.declare_parameter("det_topic", "ObjectDetections").get_parameter_value().string_value
        self._pose_topic = self.declare_parameter("pose_topic", "HeadsetPoseData").get_parameter_value().string_value
```

ros/python/py_pubsub/py_pubsub/object_detector.py
```diff
@@ -8,6 +8,7 @@
 from rclpy.qos import qos_profile_sensor_data
```
Purg, line 8: Meta-comment: we should think about creating different packages for the different nodes / grouped collections of nodes.

docker/workspace-base-dev/Dockerfile
joshanderson-kw: Yeah, this is what I had to do to get OpenGL working in the container. The trimesh scene view uses that to show the meshes.

joshanderson-kw: Although I don't think it's required if we don't use the scene viewer. https://trimsh.org/trimesh.viewer.html

ros/python/py_pubsub/py_pubsub/object_detector.py
joshanderson-kw: Oops. I was using this to debug. I took it out.

docker/workspace-base-dev/Dockerfile
Purg: Either way, I believe this belongs in the other Dockerfile since it has to do with the running of the ROS workspace, but development tools.

ros/cpp/src/ptg_datahub/src/ptg_datahub.cpp
joshanderson-kw: Good question... I thought they were needed at some point but realized they weren't pointless since the `recv` calls are blocking until we get something in the socket right now. I went ahead and removed them.

ros/python/py_pubsub/py_pubsub/spatial_mapper.py
```diff
@@ -0,0 +1,445 @@
...
+            else:
+                triangles = np.vstack([triangles,
+                                       np.array([t.vertex_indices[0],
+                                                 t.vertex_indices[1],
+                                                 t.vertex_indices[2]
+                                                ])])
+
+        if msg.removal:
+            log.debug("Got a removal!")
+            # TODO: remove from the spatial map
```
Purg, line 115: Do you intend to add removal with this PR, or is this still a punt for later? Are we prone to "leaking" mesh (accumulation w/o removal) without this removal functionality?

ros/python/py_pubsub/py_pubsub/spatial_mapper.py
```diff
@@ -0,0 +1,445 @@
...
+            if detection.source_stamp == self.poses[i].header.stamp:
+                world_matrix_1d = self.poses[i].world_matrix
+                projection_matrix_1d = self.poses[i].projection_matrix
+
+                log.debug(f"time stamps: {self.poses[i].header.stamp} {detection.source_stamp}")
+
+                # can clear out our pose list now assuming we won't get
+                # image frame detections out of order
+                self.poses = self.poses[i:]
+                break
```
Purg, line 150: Maybe this is OK for now (though I would suggest scanning in reverse as received detections are more likely to be associated with the more recent head poses), but this is a little expensive at `O(n)`.

We should use something that is more on the order of binary search (`O(log(n))`). Since we're doing strict equality here, a drop-in replacement would be using a binary search method using keys that are integer nanoseconds. I suppose that we are assuming we get a head pose for every frame that is encoded with the same time-stamp. If/when that assumption breaks then we need to use an approach that lets us find a most-recent-fallback (I did this recently in c++ for another project).

ros/angel_msgs/msg/HeadsetPoseData.msg
```diff
@@ -0,0 +1,9 @@
+#
+# Message for transporting the world matrices and projection matrices provided
+# with each frame from the HoloLens.
+#
+
+std_msgs/Header header
+
+float32[] world_matrix
+float32[] projection_matrix
```
Purg, line 9: Did you happen to consider the `PoseStamped` message from geometry_msgs? It does use quaternions though, so I'm not sure if that was the blocker that prevented it's use here (though I can find a reference for the quaternion math I needed to implement once in java to use this message type).

ros/python/py_pubsub/py_pubsub/spatial_mapper.py
joshanderson-kw: I was going to wait on it for now, but can add it if we need. The reason I hadn't added that logic yet is because I've never actually seen the `OnMeshRemoval` callback called in the Unity app.

ros/angel_msgs/msg/HeadsetPoseData.msg
joshanderson-kw: Yeah, I looked at the `geometry_msgs/Pose` and `geometry_msgs/Transform` a little bit, but wasn't exactly sure how they related to the 4x4 matrices. It seems that the 4x4 matrices are composed of a translation vector and a 3x3 rotation matrix? I wasn't sure how to go from that to a quaternion. And then if we'd end up just converting the quaternion back to the world matrix to do all of the math.

ros/python/py_pubsub/py_pubsub/spatial_mapper.py
joshanderson-kw: I went ahead and reversed the scan and added a note about implementing something more efficient.

Purg (approved):
LGTM. I think this is fine to land, which we need to be compatible with the current state of the unity app as well.

PR #30, opened 2022-03-10
Fix Mixed Reality Capture video distortion

joshanderson-kw:
Enables "render from PV camera" for mixed reality capture. By default this setting is not enabled and can cause distortion issues when MRC video capture is running.

https://docs.microsoft.com/en-us/windows/mixed-reality/develop/advanced-concepts/mixed-reality-capture-overview#render-from-the-pv-camera-opt-in)](https://docs.microsoft.com/en-us/windows/mixed-reality/develop/advanced-concepts/mixed-reality-capture-overview#render-from-the-pv-camera-opt-in

### DIFF
--- a/unity/Hello_World/Assets/MixedRealityToolkit.Generated/CustomProfiles/New MixedRealityCameraProfile.asset
+++ b/unity/Hello_World/Assets/MixedRealityToolkit.Generated/CustomProfiles/New MixedRealityCameraProfile.asset
@@ -27,7 +27,7 @@ MonoBehaviour:
     componentName: XR SDK Windows Mixed Reality Camera Settings
     priority: 0
     runtimePlatform: 9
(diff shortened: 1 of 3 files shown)


Purg:
No idea what this is changing looking at the diffs and the only way I can think to "test" this is to build it locally and install that locally built app and see if "detections look right" while recording a video. Do you have a better suggesting to check that this change has correctly propagated and portable?

Purg:
It might be good to know where the check box is so i can see if it "is checked" when I load up the project, i.e. see that the change is propagated correctly via these file changes. If i have to check the box myself, then the "changed" should be one of documentation, not specifically/just of these files. Could you pose a nav or screen clip of the box?

joshanderson-kw:
setting_update

"Render from PV Camera" should be checked with these new settings.

Purg:
Thanks!

Purg (approved):

PR #31, opened 2022-03-14
Various unity app performance improvements

joshanderson-kw:
Contains a few setting updates from recommendations found here: https://docs.microsoft.com/en-us/windows/mixed-reality/mrtk-unity/performance/perf-getting-started?view=mrtkunity-2021-05

Updates:
- Change default build quality setting from Ultra to very-low
- Disable hand mesh rendering
- Use MRTK recommended shaders


Purg:
Not opposed to this of course. What kinds of qualitative/quantitative effects have you seen with these settings changes?

Purg (approved):

PR #34, opened 2022-03-16
Dev/add tcp reconnect

joshanderson-kw:
Adds logic to catch TCP stream write fails and attempts to reconnect to the TCP client.

Logic added for PV camera stream, audio stream, and spatial mapping stream. Research mode clients will not reconnect yet.


Purg:
Hi @joshanderson-kw! I hope to be able to physically test this Monday for verification, and then we can get this in!

Purg:
Rebased on master.

PR #36, opened 2022-03-21
Dev/add task monitoring node

joshanderson-kw:
This PR adds a ROS node to track a task and publish its status. The datahub subscribes to the task updates, serializes them and then sends them to the headset. Unity app displays status in a new debug panel.


poetry.lock
```diff
@@ -159,7 +159,7 @@ python-versions = "*"
 
 [[package]]
 name = "fonttools"
-version = "4.29.1"
+version = "4.31.1"
```
Purg, line 162: This is fine in this context, but be careful about updating all dependencies when adding new ones, otherwise we get a lot of churn in dependencies, which can lead to difficulty in isolating what the cause of a "post-update-problem" is. Basically an effort to minimize parts moved.

unity/Hello_World/Packages/packages-lock.json
```diff
@@ -196,7 +196,7 @@
       "depth": 1,
       "source": "registry",
       "dependencies": {
-        "com.unity.subsystemregistration": "1.0.6"
+        "com.unity.subsystemregistration": "1.0.5"
```
Purg, line 199: Version regression? I have no idea what this is for anyway,

Purg (approved):
LGTM, some minor comments to take a look at that don't have any impact here. I'll merge shortly.

PR #40, opened 2022-04-13
Initial ARUI update message functionality

Purg:
Drafting an initial communications "standard" to push up to the pending ARUI component.


Purg:
@ovenmitt I have update this branch given today's discussion. I have taken a little liberty in how I "cleaned up" the user notification "Spatial associations" part (at the bottom of the file) to not use booleans, but lists that may contain zero or more associations for that notification. 

@ovenmitt and @joshanderson-kw, PTAL and leave any review comments that you think should block the merger of this initial format.

PR #41, opened 2022-04-14
Dev/update task monitor

joshanderson-kw:
Updates the task monitor ROS node to track time based tasks.

- Updates TaskUpdate message
- Adds TaskItem message
- Adds detection threshold for activity detector
- Adds task timer thread to task monitor node that publishes a TaskUpdate message once/sec until the timer is finished.


unity/Hello_World/Assets/Scripts/TaskManager.cs
```diff
@@ -59,15 +59,17 @@ protected void Start()
...
-
-
-        TaskUpdateMsg taskUpdateMessage = new TaskUpdateMsg(new HeaderMsg(),
-                                               "Waiting for task",
-                                               new string[0],
-                                               "N/A",
-                                               "N/A",
-                                               "N/A",
-                                               "N/A");
+        TaskUpdateMsg taskUpdateMessage = new TaskUpdateMsg(new HeaderMsg(), // ROS std header
```
Purg, line 62: I was confused for a second on seeing this in the C#. A comment here along the lines of "this is a placeholder to provide initial messaging in the UI" would be helpful.

angel_system/impls/detect_activities/pytorchvideo_slow_fast_r50.py
```diff
@@ -139,12 +147,20 @@ def detect_activities(
         # Get the predicted classes
         post_act = torch.nn.Softmax(dim=1)
         preds = post_act(preds)
-        pred_classes = preds.topk(k=5).indices
+        top_preds = preds.topk(k=5)
 
         # Map the predicted classes to the label names
+        pred_classes = top_preds.indices
```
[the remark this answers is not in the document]
joshanderson-kw: I added some comments to hopefully make it a bit more clear. The output of the model is a (1,400) tensor and that topk result is a (1,k) tensor. So that's why I am taking just the first index.

ros/angel_system_nodes/angel_system_nodes/task_monitor.py
```diff
@@ -117,11 +149,23 @@ def listener_callback(self, activity_msg):
...
 
         self.publish_task_state_message()
 
+        # Check to see if this new state has a timer associated with it
+        try:
+            if self._task.machine.states[self._task.state].timer_length > 0:
+                # Spawn a thread to track the timer and publish state messages
+                t = threading.Thread(target=self.task_timer_thread)
+                t.start()
+                self._timer_active = True
```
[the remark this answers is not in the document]
joshanderson-kw: Yeah, I wasn't sure how to approach this one really. I used the thread, rather than just doing the timer loop in the callback, because then we'd be able to receive new activity subscription callbacks and update the current activity while the timer loop is running, but there is probably a cleaner solution. All we need is to be able to send the task update messages in the timer loop while still responding to new activities.

ros/angel_system_nodes/angel_system_nodes/task_monitor.py
```diff
@@ -71,12 +93,15 @@ def __init__(self):
         self._current_activity = None
         self._next_activity = None
 
+        # Tracks whether or not a timer is currently active
+        self._timer_active = False
```
Purg, line 97: This variable, since it's used accessed/set on the main and child threads, should have an associated lock: e.g. `self._timer_lock = threading.RLock()`.

ros/angel_system_nodes/angel_system_nodes/task_monitor.py
```diff
@@ -109,6 +134,11 @@ def listener_callback(self, activity_msg):
 
         self._current_activity = current_activity
 
+        # If we are currently in a timer state, exit early since we need to wait for
+        # the timer to finish
+        if self._timer_active:
+            return
```
Purg, line 140:
```suggestion
        with self._timer_lock:
            if self._timer_active:
                return
```

ros/angel_system_nodes/angel_system_nodes/task_monitor.py
```diff
@@ -117,11 +147,23 @@ def listener_callback(self, activity_msg):
...
 
         self.publish_task_state_message()
 
+        # Check to see if this new state has a timer associated with it
+        try:
+            if self._task.machine.states[self._task.state].timer_length > 0:
+                # Spawn a thread to track the timer and publish state messages
+                t = threading.Thread(target=self.task_timer_thread)
+                t.start()
+                self._timer_active = True
```
Purg, line 162: The use of a lock will mitigate this, but we probably want to set the `_timer_active` variable *before* starting the thread (which before could theoretically set `_timer_active` before the main thread gets to the next line).
```suggestion
                with self._timer_lock:
                    self._timer_active = True
                    t = threading.Thread(target=self.task_timer_thread)
                    t.start()
```

ros/angel_system_nodes/angel_system_nodes/task_monitor.py
```diff
@@ -152,11 +210,49 @@ def publish_task_state_message(self, activity=None):
...
+        loops = self._task.machine.states[self._task.state].timer_length
+
+        # Publish a task update message once per second
+        for i in range(int(loops)):
+            self.publish_task_state_message()
+            self._task.machine.states[self._task.state].timer_length -= 1
+
+            time.sleep(1)
+
+        self._timer_active = False
```
Purg, line 240:
```suggestion
        with self._timer_lock:
            self._timer_active = False
```

Purg (approved):

PR #43, opened 2022-05-02
Dev/switch to openxr spatial awareness

joshanderson-kw:


ros/angel_system_nodes/angel_system_nodes/spatial_mapper.py
```diff
@@ -251,8 +225,10 @@ def detection_callback(self, detection):
             for p in bounds_screen_pos:
                 scaled_point = self.scale_pixel_coordinates(p)
 
-                focal_length_x = projection_matrix_2d[0][0]
-                focal_length_y = projection_matrix_2d[1][1]
+                #focal_length_x = projection_matrix[0][0]
+                focal_length_x = 1.63
+                #focal_length_y = projection_matrix[1][1]
+                focal_length_y = 2.50
```
Purg, line 231: We should add here an attention grabbing comment to highlight that this is where the hack is. Probably also at the top in the class doc-string.,

ros/angel_system_nodes/angel_system_nodes/spatial_mapper.py
```diff
@@ -351,12 +328,14 @@ def convert_pixel_coord_to_world_coord(self, world_matrix_2d,
 
         scaled_point = self.scale_pixel_coordinates(p)
 
-        focal_length_x = projection_matrix[0][0]
-        focal_length_y = projection_matrix[1][1]
-        center_x = projection_matrix[0][2]
-        center_y = projection_matrix[1][2]
+        #focal_length_x = projection_matrix[0][0]
+        focal_length_x = 1.63
```
Purg, line 332: Now that I'm seeing this set multiple times, the value should probably be set as a global constant that is used in both places. Hack commentary could below with those global defs.

unity/Hello_World/Assets/MixedRealityToolkit.Generated/CustomProfiles/New MixedRealityToolkitConfigurationProfile.asset
```diff
@@ -31,7 +31,7 @@ MonoBehaviour:
   xrsdkBoundarySystemType:
     reference: Microsoft.MixedReality.Toolkit.XRSDK.XRSDKBoundarySystem, Microsoft.MixedReality.Toolkit.Providers.XRSDK
   boundaryVisualizationProfile: {fileID: 11400000, guid: 6d28cce596b44bd3897ca86f8b24e076, type: 2}
-  enableTeleportSystem: 1
+  enableTeleportSystem: 0
```
Purg, line 34: What do these options relate to?

unity/Hello_World/Assets/MixedRealityToolkit.Generated/CustomProfiles/New MixedRealitySpatialAwarenessSystemProfile.asset
```diff
@@ -21,24 +21,10 @@ MonoBehaviour:
     priority: 0
     runtimePlatform: 8
     observerProfile: {fileID: 11400000, guid: 8be0bcd2117dd214da41ed98f0def2e3, type: 2}
-  - componentType:
```
Purg, line 24: Not really a "this PR" thing, but this file is not named very meaningfully.

unity/Hello_World/Assets/MixedRealityToolkit.Generated/CustomProfiles/New SceneUnderstandingObserverProfile.asset
```diff
@@ -17,7 +17,7 @@ MonoBehaviour:
   isStationaryObserver: 0
   observationExtents: {x: 3, y: 3, z: 3}
   observerVolumeType: 1
-  updateInterval: 5
```
Purg, line 20: Similar soft file name comment.

unity/Hello_World/Assets/Scripts/PVCameraCapture.cs
```diff
@@ -41,9 +39,10 @@ public class PVCameraCapture : MonoBehaviour
     MediaCapture mediaCapture = null;
     private MediaFrameReader frameReader = null;
     private byte[] frameData = null;
-    SpatialCoordinateSystem worldOrigin;
 #endif
 
+    public static Dictionary<Int64, TimeSpan> timeToQPCMap = new Dictionary<Int64, TimeSpan>();
```
Purg, line 44: This doesn't get used. Remove?

unity/Hello_World/Assets/Scripts/PVCameraCapture.cs
```diff
@@ -136,6 +122,11 @@ void Update()
             debugString = "";
         }
 
+#if ENABLE_WINMD_SUPPORT
+        // Update the projection matrix
+        projectionMatrix = Camera.main.projectionMatrix;
```
Purg, line 127: Is this stuff threaded? If so, should this have an associated lock/mutex?

unity/Hello_World/Assets/Scripts/PVCameraCapture.cs
```diff
@@ -280,13 +271,18 @@ unsafe private void OnFrameArrived(MediaFrameReader sender, MediaFrameArrivedEve
                     uint width = Convert.ToUInt32(originalSoftwareBitmap.PixelWidth);
                     uint height = Convert.ToUInt32(originalSoftwareBitmap.PixelHeight);
 
+                    //debugString += frame.VideoMediaFrame.CameraIntrinsics.FocalLength.ToString();
```
Purg, line 274: Remove debug lines?

unity/Hello_World/Assets/Scripts/PVCameraCapture.cs
```diff
@@ -328,23 +325,23 @@ public bool HL2TryGetCameraToWorldMatrix(MediaFrameReference frameReference, out
...
 
-        System.Numerics.Matrix4x4 cameraCoordsToUnityCoords = System.Numerics.Matrix4x4.Transpose(cameraCoordsToUnityCoordsMatrix.Value);
+        cameraCoordsToSceneCoordsMatrix = System.Numerics.Matrix4x4.Transpose(cameraCoordsToSceneCoordsMatrix);
 
         // Change from right handed coordinate system to left handed UnityEngine
-		cameraCoordsToUnityCoords.M31 *= -1f;
-		cameraCoordsToUnityCoords.M32 *= -1f;
-		cameraCoordsToUnityCoords.M33 *= -1f;
-		cameraCoordsToUnityCoords.M34 *= -1f;
+		cameraCoordsToSceneCoordsMatrix.M31 *= -1f;
```
Purg, line 339: Wasn't this not the correct RH->LH transform to the 4x4 matrix?

unity/Hello_World/Assets/Scripts/SpatialMappingCapture.cs
```diff
@@ -77,12 +94,89 @@ public class SpatialMappingCapture : MonoBehaviour, IMixedRealitySpatialAwarenes
...
+        CoreServices.SpatialAwarenessSystem.UnregisterHandler<IMixedRealitySpatialAwarenessObservationHandler<SpatialAwarenessMeshObject>>(this);
+    }
+
+    public void OnObservationAdded(MixedRealitySpatialAwarenessEventData<SpatialAwarenessMeshObject> eventData)
+    {
+        // Copy over the mesh data we need because we cannot access it outside of the main thread
+        Vector3[] vertices = eventData.SpatialObject.Filter.mesh.vertices;
+        int[] triangles = eventData.SpatialObject.Filter.mesh.triangles;
+        int id = eventData.SpatialObject.Id;
+        Thread tSendObject = new Thread(() => SendMeshObjectAddition(vertices,
```
Purg, line 159: Are we sending this to a thread because otherwise this method would block and cause undue runtime slowdown?

unity/Hello_World/Assets/Scripts/SpatialMappingCapture.cs
```diff
@@ -160,49 +259,40 @@ private void DetectionCallback(ObjectDetection3dSetMsg msg)
...
-        tSendObject.Start();
+        MeshMsg shapeMsg = new MeshMsg(meshTriangles.ToArray(), meshPoints.ToArray());
+
+        // Build and return the spatial mesh message
+        SpatialMeshMsg meshMsg = new SpatialMeshMsg(id, false, shapeMsg);
+
+        return meshMsg;
     }
 
     private void SendMeshObjectAddition(Vector3[] vertices, int[] triangles, int id)
```
Purg, line 298: We at some point need to turn a critical eye to performance, and one random thought is if this method, being thread executed, is causing some of that slow down? E.g. I don't have a feel for how often this gets called and how expensive it is. What jumps out at me is that, by threading this like we do above, we're acing in a fire-and-forget manner, meaning there there is nothing functionally stopping there being more than a managable amount of these threads doing mesh serialization.

unity/Hello_World/Assets/MixedRealityToolkit.Generated/CustomProfiles/New MixedRealityToolkitConfigurationProfile.asset
joshanderson-kw: https://docs.microsoft.com/en-us/windows/mixed-reality/mrtk-unity/features/teleport-system/teleport-system?view=mrtkunity-2021-05

I don't think it ever actually did anything since we're on the HoloLens but I figured it couldn't hurt to disable it.

unity/Hello_World/Assets/Scripts/SpatialMappingCapture.cs
joshanderson-kw: Yes, this callback is called from the main Unity thread, so the mesh sending pauses the frame rendering.

unity/Hello_World/Assets/Scripts/SpatialMappingCapture.cs
joshanderson-kw: Yeah, that is a good point. I haven't looked at how long this takes and how often it happens, but just ball parking there's usually ~20 active meshes. Spatial awareness system is configured to update once per second (we can slow this down), so this could get called up to 20 times per second.

Purg (approved):

PR #45, opened 2022-05-16
Add tmux config for bag recording

joshanderson-kw:
Add a simple tmuxinator config for recording all headset published topics to a ROS bag.

@Purg, PTAL and let me know if this is what you were thinking for the ROS bag recording.

### DIFF
--- a/tmux/record_ros_bag.yml
+++ b/tmux/record_ros_bag.yml
@@ -1,0 +1,61 @@
+#
+# Used to create ROS bags for all topics published from the HL2.
+#
+
+name: fragment-object-detection-debug
+root: <%= ENV["ANGEL_WORKSPACE_DIR"] %>
+
+# Optional tmux socket
+# socket_name: foo
+
+# Note that the pre and post options have been deprecated and will be replaced by
+# project hooks.
+
+# Project hooks
+
+# Runs on project start, always
+# on_project_start: command
+on_project_start: |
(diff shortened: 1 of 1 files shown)


tmux/record_ros_bag.yml
```diff
@@ -0,0 +1,61 @@
...
+# attach: false
+
+windows:
+  - datahub: ros2 run ros_tcp_endpoint default_server_endpoint --ros-args
+      -r __ns:=${ROS_NAMESPACE}
+      -p ROS_IP:=0.0.0.0
+  - image_converter: ros2 run angel_datahub ImageConverter --ros-args
+      -r __ns:=${ROS_NAMESPACE}
+      -p topic_input_images:=PVFramesNV12
+      -p topic_output_images:=PVFramesRGB
```
Purg, line 60: We probably don't need to run the converter even. If we do, `--all` will include both image topics, so we would be storing imagery twice. This is (A) redundant and (B) the recording might not be able to keep up depending on the host's disk IO rate. I imagine we want to keep the NV12 version which is smaller.

tmux/record_ros_bag.yml
```diff
@@ -0,0 +1,61 @@
+#
+# Used to create ROS bags for all topics published from the HL2.
```
Purg, line 2: We may want to comment that this is to be run by itself, e.g. not in combination with another tmuxinator launch. Maybe in the future we change this or create a new version that may be run in combination, but lets not venture that far yet and just document what we have.

PR #46, opened 2022-05-16
Use cv_bridge ToImageMsg to create ROS image message

joshanderson-kw:
Was a comment in a previous PR (forgot which one...) to use cv_bridge's `ToImageMsg` function to convert between a cv mat and a ROS message, rather than creating the `ImageMsg` manually.  This PR addresses that.

### DIFF
--- a/ros/angel_datahub/src/image_converter.cpp
+++ b/ros/angel_datahub/src/image_converter.cpp
@@ -101,22 +101,16 @@ ImageConverter
   cv::Mat rgb_image;
   cv::cvtColor(nv12_image, rgb_image, cv::COLOR_YUV2RGB_NV12);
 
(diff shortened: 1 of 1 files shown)


Purg (commented):
Thanks for the update! Is it correct to presume this does not change the performance of this node's bandwidth (e.g. can still handle at least 30Hz+)? If performance has not changed, LGTM.

joshanderson-kw:
> Thanks for the update! Is it correct to presume this does not change the performance of this node's bandwidth (e.g. can still handle at least 30Hz+)? If performance has not changed, LGTM.

Yeah, no difference in performance that I noticed. I can still get ~30Hz on my end with this update.

Purg (approved):

PR #48, opened 2022-05-18
Modify audio player node to use the new HeadsetAudioDataMsg

joshanderson-kw:
Depends on #47 

Node plays back the received audio data with the `SimpleAudio` package.


ros/angel_system_nodes/angel_system_nodes/audio_player.py
```diff
@@ -0,0 +1,85 @@
...
+        thread if the cumulative recording has reached more than a second and
+        there is no ongoing playback.
+        """
+        self.audio_stream.extend(msg.data)
+        self.audio_stream_duration += msg.sample_duration
+
+        if self.audio_stream_duration >= self.playback_duration and not(self.t.is_alive()):
+            # Make a copy of the current data so we can play it back
+            # while more data is accumulated
+            audio_data = copy.deepcopy(self.audio_stream)
```
Purg, line 45: Baring a complication I don't realize, I don't think you need to deep-copy here (deep copies are expensive). You are throwing away the instance we just deep copied right after this by assigning the variable a new `bytearray` instance.

ros/angel_system_nodes/angel_system_nodes/audio_player.py
```diff
@@ -0,0 +1,85 @@
...
+        self.audio_stream_duration = 0.0
+        self.playback_duration = 1.0
+
+    def listener_callback(self, msg):
+        """
+        Add the audio to the cumulative audio stream and spawns a new playback
+        thread if the cumulative recording has reached more than a second and
+        there is no ongoing playback.
+        """
+        self.audio_stream.extend(msg.data)
```
Purg, line 39: I'm not sure how ROS handles threading this, but this listener might want to use a lock to become a critical section so that message interleaving doesn't occur, probably checking that the current message is temporally after our last message.

joshanderson-kw: I added some code to check that the messages are sequential and a lock around the audio_stream. Let me know if you were thinking something else.

ros/angel_system_nodes/angel_system_nodes/audio_player.py
```diff
@@ -24,29 +23,54 @@ def __init__(self):
...
         if self.audio_stream_duration >= self.playback_duration and not(self.t.is_alive()):
             # Make a copy of the current data so we can play it back
             # while more data is accumulated
-            audio_data = copy.deepcopy(self.audio_stream)
+            audio_data = self.audio_stream
 
-            # Remove the data that we just copied from the stream
-            self.audio_stream = bytearray()
-            self.audio_stream_duration = 0.0
+            with self.audio_stream_lock:
```
Purg, line 70: I think this lock with need to protect a larger scope. Currently, `audio_data = self.audio_stream` can be subject to clobbering by modifications currently occurring in critical sections. Also, `self.prev_timestamp = msg.header.stamp` is not protected, so what is actually assigned could be subject to race conditions. A naive thing to do is protect everything from order-check to `audio_data` assignment in the `self.audio_stream_lock` context, effectively serializing all calls to `listener_callback` (`log = ...` and thread spawning don't have to be in it I don't think).

ros/angel_system_nodes/angel_system_nodes/audio_player.py
```diff
@@ -0,0 +1,109 @@
...
+            else:
+                if msg.header.stamp.sec > self.prev_timestamp.sec:
+                    message_order_valid = True
+                elif msg.header.stamp.sec == self.prev_timestamp.sec:
+                    # Seconds are same, so check nanoseconds
+                    if msg.header.stamp.nanosec > self.prev_timestamp.nanosec:
+                        message_order_valid = True
+
+            if message_order_valid:
+                with self.audio_stream_lock:
```
Purg, line 57: Nested use of the lock that we don't need anymore. This will still work because it is an `RLock` but is not needed in the new context.

joshanderson-kw: Oops.. good catch. Missed that one.

ros/angel_system_nodes/angel_system_nodes/audio_player.py
```diff
@@ -0,0 +1,108 @@
...
+                        message_order_valid = True
+
+            if message_order_valid:
+                self.audio_stream.extend(msg.data)
+                self.audio_stream_duration += msg.sample_duration
+            else:
+                log.info("Warning! Out of order messages.\n"
+                         + f"Prev: {self.prev_timestamp} \nCurr: {msg.header.stamp}")
+
+            self.prev_timestamp = msg.header.stamp
```
Purg, line 63: I just realized that if we get an out of order stamp, we're still setting the prev timestamp to the current, which may be in the past, and if so by more than one step can allow a older-than-head message to sneak in. We should likely `return` once we discover the current message is out of order.

joshanderson-kw: Good catch!

Purg (approved):

PR #53, opened 2022-06-07
Dev/change unity app name

joshanderson-kw:
Change unity app name from "Hello_World" to "ARUI"
Change main scene from "my_scene" to "arui_engineering"
Change built/deployed application name from "Hello World" to "Angel ARUI"


Purg:
I was able to test a build of this (after futzing with some settings on my end to maybe recover from the "new project who dis" syndrome with unity) with the object detection projection scenario and see the good ol' detections in the HL2.

My transmission rate and stability was unfortunately not "solved" by collecting the belkin adapter straight into the headset, though I think I saw some improved performance, but I didn't run this for a length of time that would be considered statistically significant.

What needs top be updated still are some of the README files. I have some edits to the root README file on my windows machine that I could just contribute here, but I just shut that down, so that will be a tomorrow thing. Other README files possibly need some updates as well, so I would say take a look at those for any "hello world" presence.

Purg:
@joshanderson-kw PTAL at my updates and let me know if you see any issues. Otherwise this all LGTM.

Purg (approved):

PR #55, opened 2022-06-15
ARUI merge with the angel system

ovenmitt:


joshanderson-kw:
﻿I've tested it out on my setup and I ﻿can build and load the ARUI scene without issues! 
I also added the "AngelARUI" script and "TestTapData" script to our engineering scene and see the task list is loaded and that eye tracking and voice is working. Eye tracking/voice data ROS publishing also still seem to be working.

My only question is if we need all of the pictures in the `Resource/Icons` folder, but other than that looks good! We should probably wait for @Purg to give it a look as well.

Purg:
After talking with @ovenmitt it sounded link those resources were pulled in as a "package". The total size is also pretty minimal compared to everything else pulled in, as well as I suspect these binary files are not going to be changed much. I think their presence is OK in this case at this time.

Purg:
@joshanderson-kw could you add those "integration updates" as a separate PR after this is merged?

joshanderson-kw:
> @joshanderson-kw could you add those "integration updates" you mentioned in your comment as a separate PR after this is merged?

@Purg, sure will do!

PR #56, opened 2022-06-22
Add Learn swinb transformer implementation 

joshanderson-kw:


joshanderson-kw:
@Purg, this should be ready for review now. Seems to be working with my limited coffee setup. I also tested a Kinetics 400 checkpoint and that was working too.

It loads the labels file and checkpoint from `angel_system/labels/` and `angel_system/model_checkpoints`, respectively. I didn't check those files in because I wasn't sure if we wanted to put those files somewhere else. Also wasn't sure about checking in a model checkpoint.

angel_system/impls/detect_activities/swinb/swinb_detect_activities.py
```diff
@@ -0,0 +1,211 @@
...
+        """
+        model = self._model
+        if model is None:
+            # Load the model with the checkpoint
+            model = swin_b(self._checkpoint_path, self._num_classes)
+
+            # Unfreeze head
+            for n, p in model.named_parameters():
+                if n.startswith("head"):
+                    p.requires_grad = True
```
Purg, line 103: Even though we are inferencing we want some layers to require grad? Disclaimer: I don't know much about transformers.

angel_system/impls/detect_activities/swinb/swinb_detect_activities.py
```diff
@@ -0,0 +1,211 @@
...
+
+    def get_model(self) -> "torch.nn.Module":
+        """
+        Lazy load the torch model in an idempotent manner.
+        :raises RuntimeError: Use of CUDA was requested but is not available.
+        """
+        model = self._model
+        if model is None:
+            # Load the model with the checkpoint
+            model = swin_b(self._checkpoint_path, self._num_classes)
```
Purg, line 98: This might not matter for the near term, but sometimes we are given models that are already encoded to "load" onto a GPU, and have in previous implementations used a load-call that had a parameter to load to a specific given device instead of what was encoded in the weights serialization. Is that tenable here?

angel_system/impls/detect_activities/swinb/swinb_detect_activities.py
```diff
@@ -0,0 +1,211 @@
...
+            for x in frames
+        ]
+
+        # This subsamples every n (sample rate) frames
+        frames = [
+            temporal_sampling(x, s, e, self._num_frames)
+            for x, (s, e) in zip(frames, start_end_idx)
+        ]
+
+        # Crop and random short side scale jitter
```
Purg, line 155: Probably can update this comment to not imply that we are doing random jitter here. Maybe adjust to inform that we are using `spatial_sampling` in such a way that we are not introducing random jitter by aligning min/max scale and crop-size.

angel_system/impls/detect_activities/swinb/swinb_detect_activities.py
```diff
@@ -0,0 +1,211 @@
...
+        # check for optional dependencies
+        torch_spec = importlib.util.find_spec('torch')
+        torchvision_spec = importlib.util.find_spec('torchvision')
+        if (
+            torch_spec is not None and
+            torchvision_spec is not None
+        ):
+            return True
+        else:
+            return False
```
Purg, line 211: Since the context of this package is that it requires torch/torchvision, this could just simply return True, which means we don't need to implement it as the parent definition simply returns True.

Purg (commented):
It might be good to include an example somewhere of invoking this on a known set of frames (could be stored in google drive / data.kitware.com and acquired via requests/curl).

angel_system/impls/detect_activities/swinb/swinb_detect_activities.py
```diff
@@ -0,0 +1,211 @@
...
+    def detect_activities(
+        self,
+        frame_iter: Iterable[np.ndarray]
+    ) -> Iterable[str]:
+        """
+        Formats the given iterable of frames into the required input format
+        for the swin model and then inputs them to the model for inferencing.
+        """
+        # Check that we got the right number of frames
+        assert len(frame_iter) == (self._sampling_rate * self._num_frames)
```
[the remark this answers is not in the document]
joshanderson-kw: Ok, I added the cast and some comments to the class doc-string. Hopefully that makes it a bit clearer what this function expects.

angel_system/impls/detect_activities/swinb/swinb_detect_activities.py
joshanderson-kw: Hmm I'm not sure if I understand. Is that like using `torch.load` with the `map_location` parameter?

ros/angel_system_nodes/configs/swinb_activity_det_config.json
```diff
@@ -0,0 +1,13 @@
+{
+    "type": "angel_system.impls.detect_activities.swinb.swinb_detect_activities.SwinBTransformer",
+    "angel_system.impls.detect_activities.swinb.swinb_detect_activities.SwinBTransformer": {
+        "checkpoint_path": "angel_system/model_checkpoints/swinb_coffee_tasks.pth",
+        "labels_file": "angel_system/labels/swinb_coffee_task_labels.txt",
```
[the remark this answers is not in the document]
joshanderson-kw: Could you point me to what you did on DIY AI? Or would you like to go ahead and add this?

angel_system/impls/detect_activities/swinb/swinb_detect_activities.py
```diff
@@ -0,0 +1,209 @@
...
+    def detect_activities(
+        self,
+        frame_iter: Iterable[np.ndarray]
+    ) -> Iterable[str]:
+        """
+        Formats the given iterable of frames into the required input format
+        for the swin model and then inputs them to the model for inferencing.
+        """
+        # Check that we got the right number of frames
+        assert len(list(frame_iter)) == (self._sampling_rate * self._num_frames)
```
Purg, line 131: This cast is incorrectly located: if `frame_iter` was truly an iterable, then this would "consume" it and later iterations would yield nothing. What I meant was to capture the list-ification, and then use that throughout this method body:
```suggestion
        frame_iter = list(frame_iter)
        assert len(frame_iter) == (self._sampling_rate * self._num_frames)
```

angel_system/impls/detect_activities/swinb/swinb_detect_activities.py
Purg: Yes, exactly.

ros/angel_system_nodes/configs/swinb_activity_det_config.json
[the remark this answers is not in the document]
Purg: It's probably easier if I just go ahead and add this.

PR #58, opened 2022-06-23
Use smqtk w/ configs to hydrate detectors

joshanderson-kw:


ros/angel_system_nodes/configs/default_activity_det_config.json
```diff
@@ -0,0 +1,8 @@
+{
```
Purg, line 1: The configs we use might in the future belong outside of the specific ROS node package that uses them. I don't want to bike shed this, however, so these are fine here for now.

Purg (approved):
LGTM

PR #64, opened 2022-07-01
Changed orb behavior and task list design, added support for task graph with 2 levels

ovenmitt:


unity/ARUI/Readme_ARUI.txt
```diff
@@ -0,0 +1,52 @@
...
+    {
+        {"0", "Text example MainTask 1"},
+        {"1", "Text example Subtask 1 of MainTask 1"},
+        {"1", "Text example Subtask 2 of MainTask 1"},
+	{"1", "Text example Subtask 2 of MainTask 1"},
+	{"1", "Text example Subtask 2 of MainTask 1"},
+        {"0", "Text example MainTask 2"},
+        {"0", "Text example MainTask 3"},
+	{"1", "Text example Subtask 1 of MainTask 3"},
+	{"1", "Text example Subtask 2 of MainTask 3"},
```
Purg, line 25: It's not a blocker at the moment, but there is a mixture of tabs and spaces here (and likely in other locations). For consistency in viewing across editing environments (and web viewers like here!), the general practice is to prefer spaces over tabs.

PR #65, opened 2022-07-05
Add QueryImageSize service

joshanderson-kw:
Add a ROS service so that the spatial mapper node can request the current image dimensions that are being handled.

This fixes a bug where the spatial mapper's assumed image dimensions (1920x1080) did not match the actual size (1280x720) of the images being handled, causing the boxes to appear in the wrong 3d locations.

### DIFF
--- a/ros/angel_msgs/srv/QueryImageSize.srv
+++ b/ros/angel_msgs/srv/QueryImageSize.srv
@@ -1,0 +1,4 @@
+# Query the current image dimensions
+---
+int16 image_width
+int16 image_height
--- a/ros/angel_msgs/srv/QueryImageSize.srv
+++ b/ros/angel_msgs/srv/QueryImageSize.srv
@@ -1,4 +1,6 @@
 # Query the current image dimensions
+# Note that image_width and image_height may be -1 if an image has not been
+# received by the image_converter node yet
 ---
 int16 image_width
 int16 image_height
--- a/ros/angel_system_nodes/angel_system_nodes/object_detector.py
+++ b/ros/angel_system_nodes/angel_system_nodes/object_detector.py
@@ -12,7 +12,6 @@ from smqtk_core.configuration import from_config_dict
 from smqtk_detection.interfaces.detect_image_objects import DetectImageObjects
 
 from angel_msgs.msg import ObjectDetection2dSet
-from angel_msgs.srv import QueryImageSize
 from angel_utils.conversion import from_detect_image_objects_result
 
 
@@ -49,17 +48,8 @@ class ObjectDetector(Node):
             1
         )
 
-        # Start a service to allow other nodes to get the current image size
-        self._image_size_service = self.create_service(QueryImageSize,
-                                                       "query_image_size",
-                                                       self.query_image_size_callback)
-
         self._frames_recvd = 0
         self._prev_time = -1
-
-        # These are populated from the image message and used for the image size service
-        self._image_width = -1
-        self._image_height = -1
 
         # instantiate detector from the given config
         with open(self._detector_config, "r") as f:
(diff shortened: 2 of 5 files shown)


ros/angel_system_nodes/angel_system_nodes/object_detector.py
```diff
@@ -48,9 +49,18 @@ def __init__(self):
             1
         )
 
+        # Start a service to allow other nodes to get the current image size
+        self._image_size_service = self.create_service(QueryImageSize,
```
Purg, line 53: Would it be better if this service was defined / provided by the image converter node `ros/angel_datahub/src/image_converter.cpp`? That is the "source" (that we control at least) of the image data that is being emitted into the system.

ros/angel_msgs/srv/QueryImageSize.srv
```diff
@@ -0,0 +1,4 @@
+# Query the current image dimensions
```
Purg, line 1: You establish in the code that the width and height may be `-1` if an image has not been seen yet. That should be documented here.

ros/angel_system_nodes/angel_system_nodes/spatial_mapper.py
```diff
@@ -79,14 +80,40 @@ def __init__(self):
...
         #log.set_level(10)
 
 
+    def send_image_size_request(self):
+        """
+        Sends a `QueryImageSize` request message to the QueryImageSize
+        service running in the object detector node.
+        """
+        self.req = QueryImageSize.Request()
+        self.future = self.image_size_client.call_async(self.req)
```
Purg, line 112: I don't think this future needs to be an attribute of the instance and could just be a local variable.

ros/angel_system_nodes/angel_system_nodes/spatial_mapper.py
```diff
@@ -79,14 +80,40 @@ def __init__(self):
 
         self.poses = []
 
-        # TODO: get this from a ROS message?
-        self.image_height = 1080
-        self.image_width = 1920
+        # default image dimensions used if we fail to get the current image size
```
Purg, line 83: I might think that we *don't* want defaults defined here because if we cannot get the image size then there may be issues with the system abound like there are no images being published yet. I imagine the current setup is not "succeeding" every time due to race conditions, so we are becoming half-reliant that these defaults are correct.

ros/angel_system_nodes/angel_system_nodes/object_detector.py
hdefazio: Just looping myself into this conversation

joshanderson-kw: Went ahead and moved the service there.

ros/angel_system_nodes/angel_system_nodes/spatial_mapper.py
```diff
@@ -79,14 +80,40 @@ def __init__(self):
...
+        self.image_height = 720
+        self.image_width = 1280
+
+        # attempt to get the current image size from the object detector
+        self.image_size_client = self.create_client(QueryImageSize, 'query_image_size')
+        while not self.image_size_client.wait_for_service(timeout_sec=1.0):
+            log.info("Waiting for image size service...")
+
+        r = self.send_image_size_request()
```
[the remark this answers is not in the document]
Purg: I see. We can stick with the service call in the constructor. We need that information before reliably continuing on anyway, so looping until we get it seems sensible. Maybe limiting the maximum rate we query, or timing out after a while would be nice to haves, but I don't think critical given our context: it can hard-block until the service resolves.

ros/angel_system_nodes/angel_system_nodes/spatial_mapper.py
```diff
@@ -80,38 +80,35 @@ def __init__(self):
...
-            log.warn(f"Using default image dimensions {self.image_width}x{self.image_height}")
-        else:
-            self.image_height = r.image_height
-            self.image_width = r.image_width
+                     + " Make sure the image converter node is running and receiving frames")
 
-        log = self.get_logger()
-        #log.set_level(10)
+            r = self.send_image_size_request()
+            time.sleep(1)
```
Purg, line 96: I presume you want to sleep *before* sending out the request again.

joshanderson-kw: Good catch!

Purg (approved):

PR #66, opened 2022-07-05
Add CoffeeDemoTask recipe to task monitor node

joshanderson-kw:


ros/angel_system_nodes/angel_system_nodes/task_monitor.py
```diff
@@ -52,6 +53,89 @@ def __init__(self):
...
+        # Create the transitions between steps, assuming linear steps
+        self.transitions = []
+        for idx, step in enumerate(self.steps):
+            # No transition needed for the last step
+            if idx < (len(self.steps) - 1):
+                self.transitions.append({'trigger': step['name'],
+                                         'source': step['name'],
+                                         'dest': self.steps[idx + 1]['name']})
+
+        self.machine = Machine(model=self, states=self.steps,
```
Purg, line 103: It took me a second to understand that this was modifying `self` with additional attributes and functions. Is there a quick link to the transitions docs about this for reference?

ros/angel_system_nodes/angel_system_nodes/task_monitor.py
```diff
@@ -98,11 +182,34 @@ def __init__(self):
         self._timer_lock = threading.RLock()
 
         self._activity_action_dict = {
```
Purg, line 184: Is this the mapping that links activity detector output to the task transition functions? If so, we should leave a little comment to that effect.

Purg (commented):
Looks pretty great! Just a couple of additional comments/references I'd love to see here. Thanks for getting this done so fast!

ros/angel_system_nodes/angel_system_nodes/task_monitor.py
joshanderson-kw: I added a link in the class doc string to creating the state machines. Technically, `steps` and `transitions` don't have to be `self` attributes, but I use them later to figure out what the next activity should be and what the list of steps is.

PR #67, opened 2022-07-05
Fix package.xml file isolation

Purg:
The previous method was not removing empty directories that resulted
from removing files. This caused excessive rebuild's of package
dependencies if any directory structure in the workspace was changed.
This adjusted isolation method is more explicit in retaining the target
files instead of trying to be fancy with inversion.

@joshanderson-kw Please just sanity check me here.

### DIFF
--- a/docker/workspace-build/Dockerfile
+++ b/docker/workspace-build/Dockerfile
@@ -61,7 +61,11 @@ ENV ANGEL_WORKSPACE_DIR="${ANGEL_WORKSPACE_DIR}"
 #
 FROM base AS tmp_package_files
 COPY ./ros "${ANGEL_WORKSPACE_DIR}"/src
-RUN find "${ANGEL_WORKSPACE_DIR}"/src -type f \! -name "package.xml" -print | xargs rm
+RUN mkdir "${ANGEL_WORKSPACE_DIR}"/tmp \
+ && (find "${ANGEL_WORKSPACE_DIR}"/src -type f -name 'package.xml' -print | xargs -n1 -I{} cp --parents {} "${ANGEL_WORKSPACE_DIR}"/tmp) \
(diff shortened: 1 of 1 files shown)


joshanderson-kw:
I don't think I fully understand what this is doing. Is it creating a temporary directory with the same structure as `ros` with only the package.xml files so that all of the dependencies can be installed?

joshanderson-kw:
> > I don't think I fully understand what this is doing. Is it creating a temporary directory with the same structure as `ros` with only the package.xml files so that all of the dependencies can be installed?
> 
> Yup. The previous RUN line was trying to do that by "removing everything that wasn't a package.xml file" but this method was flawed in that it left empty directories around. This is not an issue for rosdep to function, but it would invalidate the docker cache when a directory somewhere else in the code base was added/removed/changed, causing a rebuild of subsequent layers, despite the fact that nothing changed with the package.xml file layout or content. This method is more verbose, but more stable in that it quite explicitly only retains the package.xml files directly.

Ok, gotcha. Thanks! For the clarification! Looks good to me then.

joshanderson-kw (approved):

Purg:
> > > I don't think I fully understand what this is doing. Is it creating a temporary directory with the same structure as `ros` with only the package.xml files so that all of the dependencies can be installed?
> > 
> > 
> > Yup. The previous RUN line was trying to do that by "removing everything that wasn't a package.xml file" but this method was flawed in that it left empty directories around. This is not an issue for rosdep to function, but it would invalidate the docker cache when a directory somewhere else in the code base was added/removed/changed, causing a rebuild of subsequent layers, despite the fact that nothing changed with the package.xml file layout or content. This method is more verbose, but more stable in that it quite explicitly only retains the package.xml files directly.
> 
> Ok, gotcha. Thanks! For the clarification! Looks good to me then.

Thanks for the check!

PR #68, opened 2022-07-07
Add AngelARUI gameobject to arui_engineering scene

joshanderson-kw:
Add the `AngelARUI` script and `TapTestData` script to the arui_engineering scene. Modify the PTG MRTK configuration profile to use the `AngelMRTKInputSystemProfile` for speech/gesture recognition.

Engineering logging panels are still present. Wasn't sure if we wanted to remove those just yet.


Purg:
> Engineering logging panels are still present. Wasn't sure if we wanted to remove those just yet.

Fine for now, we'll likely find concrete motivation to remove then in the final pass of "play testing".

Purg (approved):

PR #69, opened 2022-07-07
Fix step order and num classes

joshanderson-kw:
Fix coffee recipe order to match 
https://docs.google.com/document/d/1jkF9Nht_-I6RFqJNLoEujPNkBzPY52jhKQqn4aCi9vY/edit#heading=h.maomtoit1ceh

Also has a few step name changes to match the activity detector label names. Tested this version of the task monitor with the latest activity detector and was able to step through the whole recipe by hacking the activity detector predictions a bit...


joshanderson-kw:
@Purg, where should I put the labels file I used? I feel like that should be checked in somewhere since it's tied so closely to the task monitor right now.

Purg:
> @Purg, where should I put the labels file I used? I feel like that should be checked in somewhere since it's tied so closely to the task monitor right now.

I'm not sure tightly coupling the activity model label list (which is activity detector implementation-specific) to the task monitor. Maybe we have this task monitor read in a config file (let's say, JSON) that maps detector labels to task steps indices. The content of the configuration file will be detector model specific and coffee task specific, but will at least be a configuration file that we can change.

joshanderson-kw:
> > @Purg, where should I put the labels file I used? I feel like that should be checked in somewhere since it's tied so closely to the task monitor right now.
> 
> I'm not sure tightly coupling the activity model label list (which is activity detector implementation-specific) to the task monitor. Maybe we have this task monitor read in a config file (let's say, JSON) that maps detector labels to task steps indices. The content of the configuration file will be detector model specific and coffee task specific, but will at least be a configuration file that we can change.

Ok, I added in a config file with the mapping from activities to step index. I also found a way with the transitions package to advance to the next step without needing the `to_state` and `activity_action` dicts with the `trigger` function, so that's why I removed those dictionaries. I tested it on my end and it looks OK (both timer steps and regular steps).

ros/angel_system_nodes/angel_system_nodes/task_monitor.py
```diff
@@ -65,7 +66,7 @@ class CoffeeDemoTask():
     Steps are listed under v1.2 here:
     https://docs.google.com/document/d/1MfbZdRS6tOGzqNSN-_22Xwmq-huk5_WdKeDSEulFOL0/edit#heading=h.l369fku95vnn
     """
-    def __init__(self):
+    def __init__(self, task_steps_file):
```
Purg, line 69: Please add a little documentation on what the format of this file is expected to be.

ros/angel_system_nodes/angel_system_nodes/task_monitor.py
```diff
@@ -75,26 +76,20 @@ def __init__(self):
...
+        # Create the list of steps
+        self.steps = []
+        for key, value in self._task_steps.items():
+            self.steps.append({'name': key.replace(' ', '_')})
+
+        # Manually add the finish step
+        self.steps.append({'name': 'Done'})
 
         # Create the transitions between steps, assuming linear steps
+        # TODO: use the step index to decide the destination step
```
Purg, line 92: It looks like this is what is happening now, so is this TODO accomplished?

tmux/task_monitor_debug.yml
```diff
@@ -71,6 +71,7 @@ windows:
       -r __ns:=${ROS_NAMESPACE}
       -p det_topic:=ActivityDetections
       -p task_update_topic:=TaskUpdates
+      -p task_steps:=${CONFIG_DIR}/default_task_steps.json
```
Purg, line 74: This config should more specifically reference that it is the MIT-LL eval 1 coffee recipe steps instead of "default".

ros/angel_system_nodes/angel_system_nodes/task_monitor.py
joshanderson-kw: Hmm I should have been more clear in the TODO. I meant for this to say that we should use the step indices in the provided task steps config file to decide the `dest` step. Right now it's just still assuming all of the steps are in linear order.

Purg (approved):

PR #70, opened 2022-07-08
Initial version of querytaskgraph service

hdefazio:
Test with:

$ tmuxinator start -p task_monitor_debug.yml

$ ros2 service call /debug/query_task_graph angel_msgs/srv/QueryTaskGraph

### DIFF
--- a/ros/angel_system_nodes/angel_system_nodes/task_monitor.py
+++ b/ros/angel_system_nodes/angel_system_nodes/task_monitor.py
@@ -1,6 +1,7 @@
 import json
 import threading
 import time
+import uuid
 from typing import Dict
 
 import numpy as np
@@ -8,9 +9,11 @@ import rclpy
 from rclpy.node import Node
 from rclpy.qos import qos_profile_sensor_data
 
-from angel_msgs.msg import ActivityDetection, TaskUpdate, TaskItem
+from angel_msgs.msg import ActivityDetection, TaskUpdate, TaskItem, TaskGraph, TaskNode
 from transitions import Machine
 import transitions
+
+from angel_msgs.srv import QueryTaskGraph
 
 
 class TeaTask():
@@ -145,6 +148,14 @@ class TaskMonitor(Node):
 
         self._task = CoffeeDemoTask(self._task_steps)
 
+        self._task_graph_service = self.create_service(
+            QueryTaskGraph,
+            "query_task_graph",
+            self.query_task_graph_callback
+        )
+
+        self._task = CoffeeDemoTask()
+
         # Represents the current state of the task
         self._current_step = self._task.state
         self._previous_step = None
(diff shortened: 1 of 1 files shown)


ros/angel_system_nodes/angel_system_nodes/task_monitor.py
```diff
@@ -282,6 +291,34 @@ def listener_callback(self, activity_msg):
...
+        and return it.
+        """
+        log = self.get_logger()
+        task_g = TaskGraph()
+
+        task_g.task_nodes = []
+        for t in self._task.steps:
+            t_node = TaskNode()
+
+            t_node.uid =  str(uuid.uuid4())
```
Purg, line 306: extra whitespace
```suggestion
            t_node.uid = str(uuid.uuid4())
```

Purg, line 306: This is going to create new UUIDs for the same tasks every time this is called. Either UUIDs should be assigned on construction, or be created "lazily" when this service is called for the first time (the former is probably simpler). UUIDs ideally might be consistent across restarts if they are based in some way on task data, e.g. the md5 checksum of the task name.

Purg: Actually, maybe one step better: since this response is static (the task graph is currently not changing during runtime), this message could be created once at construction and simply returned when this callback in invoked.

hdefazio: Re: md5 checksum, is it possible that we would have repeated tasks? Like step 5 and 10 are the same? And would we want those to have the same uid?

hdefazio: > Actually, maybe one step better: since this response is static (the task graph is currently not changing during runtime), this message could be created once at construction and simply returned when this callback is invoked.

I figured the use for this would be called once on startup (which is the other task I have to do)... if I'm following you correctly?

Purg: > Re: md5 checksum, is it possible that we would have repeated tasks? Like step 5 and 10 are the same? And would we want those to have the same uid?

Disregard me then, I don't think we need to have consistent UUIDs across processes for now so it's not really worth the effort to iron out those details with such a mechanism right now.

---

> I figured the use for this would be called once on startup (which is the other task I have to do)... if I'm following you correctly?

I presume you're also thinking about the HoloLens2 app's invocation of this which may ideally only call this "once". My comment here was just along the lines that this function from the source of truth is not idempotent. Logically it should return the same thing for the duration of the process but it currently does not.

ros/angel_system_nodes/angel_system_nodes/task_monitor.py
```diff
@@ -145,6 +148,14 @@ def __init__(self):
 
         self._task = CoffeeDemoTask(self._task_steps)
 
+        self._task_graph_service = self.create_service(
+            QueryTaskGraph,
+            "query_task_graph",
+            self.query_task_graph_callback
+        )
+
+        self._task = CoffeeDemoTask()
```
Purg, line 157: Small error in the rebase: this line was modified and master, which is present on L149 above, so remove this line.

Purg (approved):

PR #80, opened 2022-07-13
fix task strings

hdefazio:
Removes underscores from the task steps

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -58,7 +58,6 @@ services:
       - ../tmux:${ANGEL_WORKSPACE_DIR}/tmux
       - ../angel_system:${ANGEL_WORKSPACE_DIR}/angel_system
       - ../model_files:${ANGEL_WORKSPACE_DIR}/model_files
-      - ../data:${ANGEL_WORKSPACE_DIR}/data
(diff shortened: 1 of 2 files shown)


hdefazio:
@joshanderson-kw

joshanderson-kw:
Has this been tested with the activity detector? I get errors when running with the activity detector output, but it works with the keyboard input.

Should we merge #83 first and add try this one?

Purg:
> Has this been tested with the activity detector? I get errors when running with the activity detector output, but it works with the keyboard input.
> 
> Should we merge #83 first and then try this one? @Purg

Yeah, let's get that in and then rebase this on master and test again.

hdefazio:
> > Has this been tested with the activity detector? I get errors when running with the activity detector output, but it works with the keyboard input.
> > Should we merge #83 first and then try this one? @Purg
> 
> Yeah, let's get that in and then rebase this on master and test again.

This also depends on #81, can we get that merged?

hdefazio:
ok, this is tested and rebased on master

joshanderson-kw:
I still see one of my old commits here. Is that why `_task_lock` was removed?

hdefazio:
Closing in favor of #85

PR #81, opened 2022-07-13
Add keyboard shortcut to task monitor

joshanderson-kw:
Use right arrow to skip to the next step.

This requires X server running on the host. I used `sudo xhost +` to allow the connection from the container.

### DIFF
--- a/ros/angel_system_nodes/angel_system_nodes/task_monitor.py
+++ b/ros/angel_system_nodes/angel_system_nodes/task_monitor.py
@@ -205,6 +205,13 @@ class TaskMonitor(Node):
         """
         log = self.get_logger()
 
+        # If we are currently in a timer state, exit early since we need to wait for
+        # the timer to finish
+        with self._timer_lock:
+            if self._timer_active:
+                log.info(f"Waiting for timer to finish for {self._task.state}")
+                return
+
         # We are expecting a "next" step activity. We observe that this
         # activity has been performed if the confidence of that activity has
         # met or exceeded the associated confidence threshold.
@@ -229,12 +236,6 @@ class TaskMonitor(Node):
         if current_activity is None:
             # No activity matching current task.
             return
-
-        # If we are currently in a timer state, exit early since we need to wait for
-        # the timer to finish
-        with self._timer_lock:
-            if self._timer_active:
-                return
 
         # Attempt to advance to the next step
         try:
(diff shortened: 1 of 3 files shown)


ros/angel_system_nodes/angel_system_nodes/task_monitor.py
```diff
@@ -355,11 +361,39 @@ def task_timer_thread(self):
         self.publish_task_state_message()
 
 
+    def monitor_keypress(self):
+        # Collect events until released
+        with keyboard.Listener(on_press=self.on_press) as listener:
+            listener.join()
+
```
Purg, line 368: Extraneous whitespace

ros/angel_system_nodes/angel_system_nodes/task_monitor.py
```diff
@@ -355,11 +361,39 @@ def task_timer_thread(self):
         self.publish_task_state_message()
```
Purg, line 362: Extraneous whitespace

ros/angel_system_nodes/angel_system_nodes/task_monitor.py
```diff
@@ -355,11 +361,39 @@ def task_timer_thread(self):
...
+    def monitor_keypress(self):
+        # Collect events until released
+        with keyboard.Listener(on_press=self.on_press) as listener:
+            listener.join()
+
+
+    def on_press(self, key):
+        """
+        Callback function for keypress events. If the right arrow is pressed,
+        the task monitor advances to the next step.
```
Purg, line 373: It would also be a good idea if you could log an info message when the node starts with this info/instructions so that we are prompted that this is a thing (possibly at the beginning of the `monitor_keypress` function above that is threaded).

ros/angel_system_nodes/angel_system_nodes/task_monitor.py
```diff
@@ -346,7 +351,8 @@ def task_timer_thread(self):
             self._timer_active = False
 
         # Advance to the next state
-        self._task.trigger(self._task.state)
+        with self._task_lock:
+            self._task.trigger(self._task.state)
```
Purg, line 355: This might not do what we expect if a manual override keypress happens during a timer (double forward). We may want to record the current state when this method starts, and either check during/after the loop that we are still in that state before advancing (or let the trigger error with a MachineError because we are no longer in the state to progress from) and return early.

joshanderson-kw: Ok, I added a check to make sure the state doesn't change during the  timer and tested it. Also, I moved the check for if a timer is active during the ActivityDetection callback to the beginning of that function so that it doesn't try to switch activities while a timer is going.

ros/angel_system_nodes/angel_system_nodes/task_monitor.py
```diff
@@ -358,31 +363,75 @@ def task_timer_thread(self):
...
+        # Record the current state
+        curr_state = self._task.state
+
         # Publish a task update message once per second
         for i in range(int(loops)):
             self.publish_task_state_message()
-            self._task.machine.states[self._task.state].timer_length -= 1
+
+            try:
+                self._task.machine.states[self._task.state].timer_length -= 1
```
Purg, line 374: Should this also use the new `curr_state` above? Otherwise, we may be modifying a different task's timing component on accident.

joshanderson-kw: Ahh, yeah good point. That is why I added the `try`. During testing `self._task` changed and threw an error.

joshanderson-kw: Oh, I see. Hadn't thought of that. I switched it to use `curr_state`.

Purg (approved):

PR #89, opened 2022-07-15
Remove no longer used param from swinb plugin and config

Purg:


Purg:
> LGTM. Should we also go ahead and remove `det_threshold` from the pytorch video implementation?

Sure, if it is not used there anymore as well. It happened to not be immediately relevant at the time I did this. If you could contribute a commit to this branch it would be appreciated.

joshanderson-kw (approved):

PR #91, opened 2022-07-15
Add done step back

joshanderson-kw:
Removing the done step caused an error when the last activity tried to trigger.

There doesn't seem to be a good way in the ARUI to indicate that the very last step has been completed, so adding this back in for now.


Purg (approved):

Purg:
I suppose I'll tag the new master with a patch version bump

PR #92, opened 2022-07-20
added video dataset creation script

alexander-lynch:


scripts/create_custom_video_dataset.py
```diff
@@ -0,0 +1,105 @@
...
+        for i in len(os.listdir(args.label_folder)):
+            temp_df = pd.read_feather(args.label_out_path + "{}_{}.feather".format(file, i))
+            final_df = pd.concat([final_df, temp_df]).reset_index(drop=True, inplace=False)
+        print(final_df)
+        final_df.to_feather(args.label_out_path + "{}.feather".format(file))
+        
+
+
+if __name__ == "__main__":
+    ap = argparse.ArgumentParser()
```
Purg, line 94: Let's add some description to this tool so that we know what it is for and how to use it when we come back to it (module doc-string, CLI arg parser description, argument help docs, etc.)

scripts/create_custom_video_dataset.py
```diff
@@ -0,0 +1,105 @@
+import argparse
```
Purg, line 1: Given that this script is intended to be run via the CLI we could add the traditional "hashbang" line here and make it executable
```suggestion
#!/usr/bin/env python3
import argparse
```

scripts/create_custom_video_dataset.py
```diff
@@ -0,0 +1,105 @@
...
+import cv2
+import os
+import pandas as pd
+import shutil
+from PIL import Image
+
+def main(args):
```
Purg, line 8: Two newline separation (PEP8)
```suggestion
from PIL import Image


def main(args):
```

Purg (approved):

PR #93, opened 2022-07-20
Multi-modal integrations to angel_system

srihegde:
Introducing some new features:
* **Multi-stream Inputs**: Currently frames are handled along with hand pose information
* **Synchronized Listeners**: Receiving multiple streams of inputs from rosbag using message_filter package in ros2
* **Integration of MMDetectActivities Interface**: Introducing new interface for multi-modal activity detection models
* **Customizable Two-Stage Activity Detection Framework**: Added 2-stage model framework for spatio-temporal multi-modal activity recognition

TBD: Documentation of the new features + Code cleanup


angel_system/interfaces/mm_detect_activities.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+    Algorithm that detects activities from a set of input
+    frames and multi-modal auxiliary data as iterables.
+    """
+
+    @abc.abstractmethod
+    def detect_activities(
+      self,
+      frame_iter: Iterable[np.ndarray],
+      aux_data_iter: Iterable[Dict]
```
Purg, line 18: The `aux_data_iter` parameter here is pretty undefined and could be just about anything. In general, interfaces should have well-defined function signatures so that users know specifically what to provide to them, and also so that implementations know specifically what they are being provided. When things are too "loose" like this, users can easily provide incorrectly formatted inputs, and implementations can easily impose implementation-specific requirements that cause implicit coupling with using applications (e.g. no better than integrating/using the one specific implementation directly).

Due to reasons like the above, I would suggest not defining an interface specifically for your new approach and focusing on a stand-alone "implementation" to perform inference over a trained model. That implementation should be documented with specifically what it needs to perform its work. As we add more implementations/models that perform a similar input/output behavior, *then* would be a more appropriate time to create an interface as we have a better understanding of what is actually common among the implementations that we want to be hot-swappable.

poetry.lock
```diff
@@ -592,6 +592,14 @@ python-versions = ">=2.6, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
 [package.extras]
 test = ["ipaddress", "mock", "unittest2", "enum34", "pywin32", "wmi"]
 
+[[package]]
```
Purg, line 595: I'm not sure why the poetry lock file was updated here as there have been no updates to dependencies. Also the trimesh updated version now doesn't match the stored hashes which I presume will cause errors when attempting a poetry install.

angel_system/impls/detect_activities/two_stage/two_stage_detect_activities.py
```diff
@@ -0,0 +1,186 @@
...
+            frames = frames.cuda()
+            for k in aux_data:
+                aux_data[k] = aux_data[k].cuda()
+
+        # Predict!
+        with torch.no_grad():
+            preds = self._model(frames, aux_data)
+
+        # Get the top predicted classes
+        post_act = torch.nn.Softmax(dim=1)
```
Purg, line 152: Could this be created once in the constructor?

angel_system/impls/detect_activities/two_stage/two_stage_detect_activities.py
```diff
@@ -0,0 +1,186 @@
...
+
+        # Filter out any detections below the threshold
+        predictions = []
+        pred_values = top_preds.values[0]
+        for idx, p in enumerate(pred_class_names):
+            if (pred_values[idx] > self._det_threshold):
+                predictions.append(p)
+
+        # pdb.set_trace()
+        return predictions
```
Purg, line 170: We revised the other activity detector interface to not filter on a confidence threshold and instead return all class labels mapped to their confidence. This is so that downstream analytics may utilize confidence values in their own way. This should similarly return confidences with the predicted labels as well.

angel_system/impls/detect_activities/two_stage/two_stage_detect_activities.py
```diff
@@ -0,0 +1,186 @@
...
+            "checkpoint_path": self._checkpoint_path,
+            "num_frames": self._num_frames,
+            "sampling_rate": self._sampling_rate,
+            "labels_file": self._labels_file,
+        }
+
+    @classmethod
+    def is_usable(cls) -> bool:
+        # Only torch/torchvision required
+        return True
```
Purg, line 186: This does not need to be implemented here as the super-class implementation already just returns true.

ros/angel_system_nodes/angel_system_nodes/mm_activity_detector.py
```diff
@@ -0,0 +1,171 @@
...
+BRIDGE = CvBridge()
+    
+
+class MMActivityDetector(Node):
+
+    def __init__(self):
+        super().__init__(self.__class__.__name__)
+
+        self._image_topic = self.declare_parameter("image_topic", "/debug/PVFrames").get_parameter_value().string_value
+        self._hand_topic = self.declare_parameter("hand_pose_topic", "/debug/HandJointPoseData").get_parameter_value().string_value
```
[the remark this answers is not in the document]
srihegde: Updated.

angel_system/impls/detect_activities/two_stage/two_stage_detect_activities.py
```diff
@@ -0,0 +1,186 @@
...
+            if self._torch_device != 'cpu':
+                if torch.cuda.is_available():
+                    model_device = torch.device(device=self._torch_device)
+                    model = model.to(device=model_device)
+                else:
+                    raise RuntimeError(
+                        "Use of CUDA requested but not available."
+                    )
+            else:
+                model_device = torch.device(self._torch_device)
```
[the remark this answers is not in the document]
srihegde: Looks good. This was probably the result of copying the existing code. Will update in the next PR

angel_system/impls/detect_activities/two_stage/two_stage_detect_activities.py
```diff
@@ -0,0 +1,186 @@
...
+        frames = [self.transform(f) for f in frame_iter]
+        frames = torch.stack(frames)
+        for k in aux_data:
+            aux_data[k] = torch.Tensor(aux_data[k])
+
+        # Move the inputs to the GPU if necessary
+        if self._model.cuda:
+            frames = frames.cuda()
+            for k in aux_data:
+                aux_data[k] = aux_data[k].cuda()
```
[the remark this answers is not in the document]
srihegde: Done!

ros/angel_system_nodes/angel_system_nodes/mm_activity_detector.py
```diff
@@ -0,0 +1,171 @@
...
+        log.info(f"Use cuda? {self._use_cuda}")
+        log.info(f"Frames per detection: {self._frames_per_det}")
+        log.info(f"Detector config: {self._detector_config}")
+
+        self.subscription_list = []
+        # Image subscription
+        self.subscription_list.append(mf.Subscriber(self, Image, self._image_topic))
+        # Hand pose subscription
+        self.subscription_list.append(mf.Subscriber(self, HandJointPosesUpdate, self._hand_topic))
+        self.time_sync = mf.TimeSynchronizer(
```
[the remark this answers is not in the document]
srihegde: If we want to use `ApproximateTimeSynchronizer` then we will need to set a `slop` parameter that is the delay margin. Not sure what the correct margin is and this may also need extra checking like - single right/left hand within the slop interval and no more than 2 hands in the slop interval etc.

angel_system/interfaces/detect_activities.py
```diff
@@ -30,4 +30,4 @@ def __call__(
         """
         Calls `detect_activities() with the given iterable set of frames.`
         """
-        return self.detect_objects(frame_iter)
+        return self.detect_activities(frame_iter)
```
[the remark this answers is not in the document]
srihegde: Thanks!

angel_system/impls/detect_activities/two_stage/two_stage_detect_activities.py
srihegde: Yes, we can. But I removed class and am directly using the softmax functional from pytorchnn.functional (as `post_act` is not used anywhere else).

angel_system/impls/detect_activities/two_stage/two_stage_detect_activities.py
srihegde: Ok. Removed.

angel_system/impls/detect_activities/two_stage/two_stage_detect_activities.py
srihegde: Ok. So you mean, 
```suggestion
        # Map class names with confidence
        pred_values = top_preds.values[0].tolist()
        predictions = dict(zip(pred_class_names, pred_values))

        # pdb.set_trace()
        return predictions
```
return

Purg: Yup, lgtm.

angel_system/impls/detect_activities/two_stage/two_stage_detect_activities.py
```diff
@@ -0,0 +1,186 @@
...
+        frame_iter: Iterable[np.ndarray],
+        aux_data_iter: Iterable[Dict]
+    ) -> Iterable[str]:
+        """
+        Formats the given iterable of frames into the required input format
+        for the swin model and then inputs them to the model for inferencing.
+        """
+        # Check that we got the right number of frames
+        frame_iter = list(frame_iter)
+        aux_data = dict(aux_data_iter)
```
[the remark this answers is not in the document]
srihegde: `aux_data` is a dictionary of iterables. For eg. in this case, we have,
```
aux_data = dict(
       lhand = [...],    # List of left hand's 21 keypoint coordinates (in 3D) (dimension: num_of_frames * 63)
       rhand = [...],    # Same as above but for right hand
) 
```


Some of the things are a little confusing to me here. What do you expect the structure of `frame_iter` to be? I get that its a list of frames, but how are you accounting for the batch sizes here?

angel_system/interfaces/mm_detect_activities.py
srihegde: I have added some info to [this](https://github.com/PTG-Kitware/angel_system/pull/93#discussion_r926730668) comment. Will look into this comment once the linked comment is clarified.

angel_system/impls/detect_activities/two_stage/two_stage_detect_activities.py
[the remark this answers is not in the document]
Purg: My comment is related to the type annotation for the parameter. `Iterable[Dict]` would be something like `[{}, {}, {}, ...]`. It sounds like you more expect a `Dict[str, Iterable[np.ndarray]]`.

> What do you expect the structure of frame_iter to be? I get that it's a list of frames, but how are you accounting for the batch sizes here?

For the previous interface that Josh authored (I presume based on the object detection interface), the `frame_iter` was the volume of frames to predict over, e.g. `frame_iter` input, a single prediction dictionary output. I would probably suggest that its meaning here is the same. How this related to batch size should be up to the algorithm, e.g. the algorithm should provide the mechanism to appropriately fit the input volume to what the model requires (e.g. more requires a very specific number of frames? --> perform subsampling). Relating to my comment about not going for an interface at this stage in the game, you could also be more implementation-specific in what `frame_iter` mean, including changing the name and type annotation, e.g. if this implementation specifically requires that there be 64 frames input, then we should document that as an input requirement.

angel_system/interfaces/mm_detect_activities.py
Purg: My comment here is that we should not have an interface at this time. We don't have multiple implementations to switch between, and we don't know the practical generalization of this based on practical application. We just, at a minimum, want to be able to simply perform forward inference passes on the model that is being trained.

angel_system/impls/detect_activities/two_stage/two_stage_detect_activities.py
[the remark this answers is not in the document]
srihegde: By this explanation, do you mean I change the signature of `aux_data_iter` from `aux_data_iter: Iterable[Dict]` to `aux_data_iter: Dict[str, Iterable[np.ndarray]]`?

angel_system/interfaces/mm_detect_activities.py
srihegde: Cool. I have removed the `MMDetectActivities` interface for now.

angel_system/impls/detect_activities/two_stage/two_stage_detect_activities.py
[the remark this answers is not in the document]
srihegde: Got it. Updated the same in the `MMDetectActivities` interface. For now I have kept it in the repo anticipating future requirements.

angel_system/interfaces/mm_detect_activities.py
Purg: Is this overridden by the previous comment that states you kept the interface around? I assume yes since I still see the interface file in the diff.

angel_system/impls/detect_activities/two_stage/two_stage.py
```diff
@@ -0,0 +1,39 @@
...
+    Args: TBD
+    """
+
+    def __init__(self, checkpoint: str, num_classes: int):
+        super().__init__()
+
+        self.fcn = SpatialFCNModule('resnext')
+        self.temporal = RULSTM(num_classes, hidden=128, dropout=0, depth=3)
+
+    def forward(self, data, aux):
```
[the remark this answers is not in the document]
Purg: This didn't receive a documentation update. @alexander-lynch Could you trace what the output of this is and leave a comment to inform the reader?

angel_system/interfaces/mm_detect_activities.py
srihegde: Oops. Yes I have kept the interface (anticipating future use) but am not using it currently.

angel_system/impls/detect_activities/two_stage/two_stage.py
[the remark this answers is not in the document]
srihegde: Oops (again). I just updated it in the training repo. Let me copy over here.

Purg (approved):

PR #94, opened 2022-07-22
Add annotation event monitor node

joshanderson-kw:
New ROS node monitors keyboard to generate annotation event messages.

Updated bag extractor to write AnnotationEvent messages to a file and did some reformatting.

Can map foot pedal buttons to left/right arrow keys to use this node hands free.

### DIFF
--- a/ros/angel_system_nodes/angel_system_nodes/annotation_event_monitor.py
+++ b/ros/angel_system_nodes/angel_system_nodes/annotation_event_monitor.py
@@ -1,0 +1,112 @@
+import threading
+from threading import Lock
+
+from pynput import keyboard
+import rclpy
+from rclpy.node import Node
+
+from angel_msgs.msg import AnnotationEvent
+
+
+class AnnotationEventMonitor(Node):
+    """
+    ROS node that monitors the keyboard to generate AnnotationEvent
+    messages. Event messages for annotation start and stop are generated
+    when the right arrow key is pressed. Event messages for error start
+    and stop are are generated when the left arrow key is pressed.
+    """
+
+    def __init__(self):
+        super().__init__(self.__class__.__name__)
+
+        self._annotation_event_topic = self.declare_parameter(
+            "annotation_event_topic", "AnnotationEvents"
+        ).get_parameter_value().string_value
+
+        # Initialize ROS hooks
+        self._publisher = self.create_publisher(
+            AnnotationEvent,
+            self._annotation_event_topic,
+            1
+        )
+
+        # Whether or not an annotation is currently ongoing
+        self._annotation_active = False
+
+        # Whether or not an error is currently ongoing
+        self._error_active = False
+
+        self._keyboard_lock = Lock()
+
+    def monitor_keypress(self):
+        log = self.get_logger()
+        log.info("Starting keyboard monitor")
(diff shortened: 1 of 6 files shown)


ros/angel_utils/scripts/bag_extractor.py
```diff
@@ -82,14 +118,20 @@ def __init__(self):
...
+        if self.extract_annotation_event_data:
+            self.msg_type_to_handler_map[
+                AnnotationEvent
+            ] = self.handle_annotation_event_msg
 
         # Top level data folder
         self.num_total_msgs = 0
         self.data_folder = self.bag_path + "_extracted/"
-        if not(os.path.exists(self.data_folder)):
+        if not (os.path.exists(self.data_folder)):
```
Purg, line 134: Given the reformatting, the parentheses around the os function are extraneous. Same for similar instances below.
```suggestion
        if not os.path.exists(self.data_folder):
```

ros/angel_utils/scripts/bag_extractor.py
```diff
@@ -332,7 +420,9 @@ def handle_image_msg(self, msg: Image) -> None:
         # Save image to disk
         timestamp_str = f"{msg.header.stamp.sec:011d}_{msg.header.stamp.nanosec:09d}"
         file_name = (
-            f"{self.image_folder}frame_{self.num_image_msgs:05d}_" + timestamp_str + ".png"
+            f"{self.image_folder}frame_{self.num_image_msgs:05d}_"
+            + timestamp_str
+            + ".png"
```
Purg, line 425: Why not include the `timestamp_str` in the format string?
```suggestion
            f"{self.image_folder}frame_{self.num_image_msgs:05d}_{timestamp_str}.png"
```

ros/angel_system_nodes/angel_system_nodes/annotation_event_monitor.py
```diff
@@ -0,0 +1,112 @@
...
+
+from angel_msgs.msg import AnnotationEvent
+
+
+class AnnotationEventMonitor(Node):
+    """
+    ROS node that monitors the keyboard to generate AnnotationEvent
+    messages. Event messages for annotation start and stop are generated
+    when the right arrow key is pressed. Event messages for error start
+    and stop are are generated when the left arrow key is pressed.
```
Purg, line 16: I know we're not also running the task monitor node to technically conflict, but there is a little mental overlap being created by using the same buttons. I'm entering a little into bikesheding territory here, but I might recommend monitoring some different keys for this.

Purg (approved):

ros/angel_system_nodes/angel_system_nodes/annotation_event_monitor.py
joshanderson-kw: Ok, good idea. I switched it to use the up/down arrow keys instead.

Purg: Thanks for the change. In an idea world, we should have a document overviewing the UI/UX of various components, which would include what keys are recognized by what nodes, so as to be a reference for users as well as development to inform what keys are "reserved" or not. I don't think this is a priority, but something to keep in the back of your mind. This would also somewhat coincide with breaking the README into more modular documentation that hasn't happened yet.

PR #95, opened 2022-07-27
Add object descriptor node

joshanderson-kw:
Add `ObjectDescriptors` message and new object descriptor ROS node.

Currently uses pretrained torchvision ResNet. 

I didn't find a super straightforward way to get the output of an arbitrary layer in a torch model, so I ended up creating a custom class that overrides the forward call to return the layers we want (the backbone of the ResNet model). @Purg, maybe you know of a better way.


ros/angel_system_nodes/setup.py
```diff
@@ -28,7 +28,8 @@
                 'spatial_mapper = angel_system_nodes.spatial_mapper:main',
                 'activity_detector = angel_system_nodes.activity_detector:main',
                 'task_monitor = angel_system_nodes.task_monitor:main',
-                'feedback_generator = angel_system_nodes.feedback_generator:main'
+                'feedback_generator = angel_system_nodes.feedback_generator:main',
+                'descriptor_generator = angel_system_nodes.object_descriptor_generator:main'
```
Purg, line 32: Might as well end the line with a comma so that we don't have to remember to modify the line when we add something new in the future.

ros/angel_system_nodes/angel_system_nodes/object_descriptor_generator.py
```diff
@@ -0,0 +1,215 @@
...
+            ObjectDescriptors,
+            self._desc_topic,
+            1
+        )
+
+        self.transforms = transforms.Compose([
+            transforms.ToTensor(),
+        ])
+
+    def get_model(self) -> "torch.nn.Module":
```
Purg, line 127: Doesn't need to be in quotes
```suggestion
    def get_model(self) -> torch.nn.Module:
```

ros/angel_system_nodes/angel_system_nodes/object_descriptor_generator.py
```diff
@@ -0,0 +1,215 @@
...
+        Modified version of thetorchvision `GeneralizedRCNN` forward call().
+        """
+        original_image_sizes: List[Tuple[int, int]] = []
+        for img in images:
+            val = img.shape[-2:]
+            torch._assert(
+                len(val) == 2,
+                f"expecting the last two dimensions of the Tensor to be H and W instead got {img.shape[-2:]}",
+            )
+            original_image_sizes.append((val[0], val[1]))
```
Purg, line 74: This doesn't seem to get used in this context?

ros/angel_system_nodes/angel_system_nodes/object_descriptor_generator.py
```diff
@@ -0,0 +1,215 @@
...
+        )
+        self._publisher = self.create_publisher(
+            ObjectDescriptors,
+            self._desc_topic,
+            1
+        )
+
+        self.transforms = transforms.Compose([
+            transforms.ToTensor(),
+        ])
```
Purg, line 125: Since this is just one thing it could just be 
```suggestion
        self.transforms = transforms.ToTensor()
```

ros/angel_system_nodes/angel_system_nodes/object_descriptor_generator.py
```diff
@@ -0,0 +1,215 @@
...
+
+        # Convert ROS Image message to CV2
+        rgb_image = BRIDGE.imgmsg_to_cv2(image, desired_encoding="rgb8")
+
+        # Convert to tensor
+        image_tensor = self.transforms(rgb_image)
+        image_tensor = image_tensor.unsqueeze(0)
+
+        # Send to model
+        output = model(image_tensor)
```
Purg, line 177: I presume we want no-grad around this (or decorated on this method)
```suggestion
        with torch.no_grad():
            output = model(image_tensor)
```

ros/angel_system_nodes/angel_system_nodes/object_descriptor_generator.py
```diff
@@ -0,0 +1,215 @@
...
+
+        if self._feature_layer == "head":
+            # Use the object scores as the features
+            output = output[0]['scores']
```
Purg, line 181: Just doing some rough testing in a terminal, it seems that the specific feature-layer parameterization will always return values, but this may be empty. I tested by inputting a random image, so part of the default functionality found "no detections" and thus output nothing.

ros/angel_system_nodes/angel_system_nodes/object_descriptor_generator.py
```diff
@@ -0,0 +1,215 @@
...
+            torch._assert(
+                len(val) == 2,
+                f"expecting the last two dimensions of the Tensor to be H and W instead got {img.shape[-2:]}",
+            )
+            original_image_sizes.append((val[0], val[1]))
+
+        images, targets = self.transform(images, targets)
+
+        features = self.backbone(images.tensors)
+        return features
```
Purg, line 79: I would want to get feedback from @cfunk1210 and/or Sri about what descriptors we want from the object detector. E.g. do we instead want a descriptor per proposal/detection, as opposed to this set of generic backbone features?

ros/angel_system_nodes/angel_system_nodes/object_descriptor_generator.py
```diff
@@ -0,0 +1,215 @@
...
+BRIDGE = CvBridge()
+
+
+class ResNetDescriptors(FasterRCNN):
+    """
+    Custom ResNet class to override the forward method call
+    of the torch `FasterRCNN` class.
+    """
+
+    def __init__(self, *args, **kwargs):
```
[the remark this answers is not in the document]
joshanderson-kw: No, that was leftover from something else. I went ahead and removed them.

ros/angel_system_nodes/angel_system_nodes/object_descriptor_generator.py
joshanderson-kw: Oh, true, good point. It looks like in that case the descriptor list in the message is just empty and dimensions is set to `[0]`.

Is that OK or should we do something different for this case?

Purg: Answering this sort of ties into the comment below about "what is an object detection descriptor", so let's resolve that. I'll start a chat thread.

Purg:
Pending update to pull out the region-proposal specific descriptors. There is a plan, we just need to do it. See relevant Monday comment.

joshanderson-kw:
@Purg, I think this is ready to review now. I ended up using a different model entirely (the one that Dawei used). Those files are in `angel_system/fasterrcnn` and copied from his repo with just a few minor tweaks to make it run in our setup. I have the model weights file locally, but not sure what the best way to store it is.

ros/angel_system_nodes/angel_system_nodes/object_descriptor_generator.py
```diff
@@ -0,0 +1,323 @@
...
+            im_info = Variable(im_info)
+            num_boxes = Variable(num_boxes)
+            gt_boxes = Variable(gt_boxes)
+
+        if len(im_in.shape) == 2:
+            im_in = im_in[:,:,np.newaxis]
+            im_in = np.concatenate((im_in,im_in,im_in), axis=2)
+        im = im_in
+
+        blobs, im_scales = _get_image_blob(im)
```
Purg, line 190: `_get_image_blob` is documented saying that it expects a "BGR" image, but I expect that at this point the image input is in RGB (see L125).

ros/angel_system_nodes/angel_system_nodes/object_descriptor_generator.py
```diff
@@ -0,0 +1,323 @@
...
+        model = self.get_model()
+
+        # Preprocess image
+        im_in = np.array(BRIDGE.imgmsg_to_cv2(image, desired_encoding="rgb8"))
+        im_data, im_info, gt_boxes, num_boxes, im_scales = self.preprocess_image(im_in)
+
+        # Send to model
+        rois, cls_prob, \
+        _, _, _, _, _, _, \
+        pooled_feat = model(im_data, im_info, gt_boxes, num_boxes, pool_feat=True)
```
Purg, line 131: Should this be within a `with torch.no_grad():` context?

ros/angel_system_nodes/angel_system_nodes/object_descriptor_generator.py
```diff
@@ -0,0 +1,323 @@
...
+                    boxs_inds = np.expand_dims(boxs_inds, 0)
+                cls_boxes = (
+                    torch.tensor(boxs_inds[:, j * 4:(j + 1) * 4])
+                    .to(device=self._torch_device)
+                )
+
+                cls_dets = torch.cat((cls_boxes, cls_scores.unsqueeze(1)), 1)
+                cls_dets = cls_dets[order]
+                keep = nms(cls_boxes[order, :], cls_scores[order], 0.3)
+                cls_dets = cls_dets[keep.view(-1).long()]
```
Purg, line 247: `cls_dets` doesn't seem to be used.

Purg (commented):
Mentioned offline, but `angel_system/fasterrcnn/` is missing an `__init__.py`.

If this fasterrcnn implementation was taken or adapted from another author, we need to reference them here in some manner.

ros/angel_msgs/msg/ObjectDetection2dSet.msg
```diff
@@ -37,3 +37,10 @@ float32[] bottom
 # This is a flattened 2D row-major matrix with shape:
 #   [num_detections, len(label_vec)]
 float64[] label_confidences
+
+# Original descriptor dimensions. For example, if the descriptor's dimensions
+# are (256, 16, 16) then this array should be set to [256, 16, 16].
+int32[] descriptor_dims
```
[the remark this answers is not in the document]
joshanderson-kw: Yeah, you are right that it could be just a single int. I went ahead and simplified it.

ros/angel_system_nodes/angel_system_nodes/object_descriptor_generator.py
joshanderson-kw: Good catch!

ros/angel_system_nodes/angel_system_nodes/object_descriptor_generator.py
```diff
@@ -0,0 +1,323 @@
...
+        )
+        self._object_vocabulary = (
+            self.declare_parameter("object_vocab_list",
+                                   "/angel_workspace/angel_system/fasterrcnn/objects_vocab.txt")
+            .get_parameter_value()
+            .string_value
+        )
+        self._model_checkpoint = (
+            self.declare_parameter("model_checkpoint",
+                                   "/angel_workspace/model_files/faster_rcnn_res101_vg.pth")
```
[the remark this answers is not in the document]
joshanderson-kw: I think I was able to upload the files... let me know if it works for you.

ros/angel_msgs/msg/ObjectDetection2dSet.msg
```diff
@@ -37,3 +37,9 @@ float32[] bottom
 # This is a flattened 2D row-major matrix with shape:
 #   [num_detections, len(label_vec)]
 float64[] label_confidences
+
+# Vector length of each descriptor.
```
Purg, line 41: Nitpick: might be good to clarify that the length of `descriptors` should be `num_detections * descriptor_dims`.

ros/angel_msgs/msg/ObjectDetection2dSet.msg
```diff
@@ -37,3 +37,9 @@ float32[] bottom
...
+int32 descriptor_dims
```
Purg, line 42: Nitpick: should this be singular? This is describing just the consistent length dimension of a single descriptor vector.

ros/angel_system_nodes/angel_system_nodes/object_descriptor_generator.py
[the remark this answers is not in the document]
Purg: I was able to pull the files with the correct sha512 and see them in the appropriate locate.

Purg (commented):
Looking good, just a couple nitpicks, I want to reproduce functionality on my end.

Purg (approved):

PR #101, opened 2022-08-15
Additional mm integration

alexander-lynch:
I had to add the src folder from sri's ptg-activity-recognition repo in order to load the checkpoint file. The main code changes occur in angel_system/impls/detect_activities/two_stage/two_stage.py where I added some lines to add in the checkpoint. I also added tmux/multimodal_activity_recognition_debug.yml as a tmux script for the multimodal activity recognition development on the hololens.


angel_system/impls/detect_activities/swinb/swinb_detect_activities.py
```diff
@@ -99,8 +99,16 @@ def get_model(self) -> torch.nn.Module:
...
+            if self._torch_device != 'cpu':
+                if torch.cuda.is_available():
+                    model_device = torch.device(device=self._torch_device)
+                    model = model.to(device=model_device)
+                else:
+                    raise RuntimeError(
+                        "Use of CUDA requested but not available."
+                    )
+            else:
+                model_device = torch.device(self._torch_device)
```
Purg, line 111: I had mentioned to Sri that this more complicated block shouldn't be necessary. If we try to send the model to an invalid device, we should get an appropriate error. There are also other devices than just CPU and CUDA that I don't see why we specifically need to cut out of the picture (not that we're likely to use them). Was this change based on an experience you had?

tmux/multimodal_activity_recognition_debug.yml
```diff
@@ -0,0 +1,98 @@
+#
+# Fragment of the system that includes data capture, object detection, and
+# debug visualization of the detections in the image space.
+#
+
+name: 20220719-darpa-sitevisit-coffee-demo
```
Purg, line 6: This config and the `activity_detection_debug.yml` probably serve the same purpose. Let's consolidate them instead of having two different configs that do basically the same thing.

angel_system/impls/detect_activities/two_stage/two_stage.py
```diff
@@ -23,13 +25,20 @@ def __init__(self, checkpoint: str, num_classes: int):
...
+        
         self.fcn.eval()
+        
         # Load checkpoint
-        self.temporal.load_state_dict(torch.load(checkpoint))
+        temp_check = dict()
+        for key in check['state_dict'].keys():
+            if 'temporal' in key:
+                new_key = key.split('temporal.')[1]
+                temp_check[new_key] = check['state_dict'][key]
```
Purg, line 40: Are there some additional comments we can add that explain this additional logic?

angel_system/impls/detect_activities/two_stage/two_stage.py
```diff
@@ -4,6 +4,8 @@
 from torch import nn
 # from torchmetrics import MaxMetric
 # from torchmetrics.classification.accuracy import Accuracy
+import sys
+sys.path.insert(0, '/angel_workspace/angel_system/impls/detect_activities/two_stage')
```
Purg, line 8: We should endeavor to not do this. I presume this was added as a quick fix to resolve import errors in submodules. We should instead fix those submodule imports.

angel_system/impls/detect_activities/two_stage/src/vendor/__init__.py
```diff
@@ -0,0 +1 @@
+# use this folder for storing third party code that cannot be installed using pip/conda
```
Purg, line 1: If there is nothing here, we should remove this submodule.

tmux/multimodal_activity_recognition_debug.yml
alexander-lynch: One is for the learn activity recognition and the other is for sri's multimodel model. Is it useful to have both of them?

Purg: I don't think it is, because the difference is only a single node that we can swap (we can leave the unused on commented out for now). In an ideal future, we have an index in the documentation of all the nodes that we have to inform someone on what can be added/removed here.

angel_system/impls/detect_activities/two_stage/two_stage.py
```diff
@@ -28,20 +24,10 @@ def __init__(self, checkpoint: str, num_classes: int):
...
-        temp_check = dict()
-
-        for key in check['state_dict'].keys():
-            #extract just the temporal weights from the checkpoint
-            #fcn doesn't change, so no need to load in weights for it
-            if 'temporal' in key:
-                new_key = key.split('temporal.')[1]
-                temp_check[new_key] = check['state_dict'][key]
-        self.temporal.load_state_dict(temp_check)
+        self.temporal.load_state_dict(torch.load(checkpoint))
```
Purg, line 30: We should comment here that we expect to receive weights for the temporal layers only.

angel_system/impls/detect_activities/two_stage/src/models/tbn_lightning.py
```diff
@@ -0,0 +1,186 @@
...
+import pdb
+from typing import Any, Dict, List
+
+import torch
+from pytorch_lightning import LightningModule
+from torchmetrics import MaxMetric
+from torchmetrics.classification.accuracy import Accuracy
+
+
+class TBNLightning(LightningModule):
```
Purg, line 16: I don't think this class get's used in the inference side of the world from what I can find. Otherwise, we should remove this, which also removes the use of a dependency (`pytorch_lightning`) that is not listed as a dependency of this package.

angel_system/impls/detect_activities/two_stage/src/models/unified_ho_module.py
```diff
@@ -0,0 +1,181 @@
+from typing import Any, Dict, List
+
+import torch
+from pytorch_lightning import LightningModule
+from torchmetrics import MaxMetric
+from torchmetrics.classification.accuracy import Accuracy
+
+
+class UnifiedHOModule(LightningModule):
```
Purg, line 9: Another pytorch lightning use and class that doesn't seem  to be used.

angel_system/impls/detect_activities/two_stage/src/models/components/tbn_deps/ops/__init__.py
```diff
@@ -0,0 +1 @@
+from ops.basic_ops import *
```
Purg, line 1: Is this supposed to be a local import?
```suggestion
from .basic_ops import *
```

angel_system/impls/detect_activities/two_stage/src/models/components/tbn_deps/ops/utils.py
```diff
@@ -0,0 +1,36 @@
+import numpy as np
+import torch
+from sklearn.metrics import confusion_matrix
```
Purg, line 3: If this module is used, scikit-learn is not added as a dep in `pyproject.toml`.

angel_system/impls/detect_activities/two_stage/src/models/components/unified_fcn.py
```diff
@@ -0,0 +1,78 @@
+from collections import OrderedDict
+from multiprocessing import Pool
+from typing import Dict, Tuple
+
+import torch
+from torch import nn
+from torchvision.models import resnext50_32x4d  # ,convnext_tiny
+
+# from torchvision.models.feature_extraction import create_feature_extractor
```
Purg, line 9: Remove dead code.

angel_system/impls/detect_activities/two_stage/src/models/components/unified_fcn_unfinished.py
```diff
@@ -0,0 +1,223 @@
+# """
```
Purg, line 1: This whole file is commented out, so it could probably be removed?

Purg:
Rebasing on current master.

Purg (approved):

PR #102, opened 2022-08-16
addition of footpedal parsing script

alexander-lynch:


scripts/parse_footpedal_anns.py
```diff
@@ -14,24 +13,27 @@ def main(args):
...
     :param args.ann_path: Absolute path to the annotation_event_data.txt file as is output by the exploded ros bag
                           script
-    :param args.rosbag_name: A string containing the name of the rosbag that is being processed
+    :param args.rosbag_name: A string containing the name of the rosbag that is being processed. This is used to 
+                             create a relative filepath to the exploded rosbag image folder for evaluation.
     :param args.activity: Only used when parsing directly to the PTG labels, this is for when a recording only contains
-                     a single activity being performed repeatedly. the parameter is a string of the name of the activity
-    :param args.ptg_out_file: The absolute path of the feather file containing the PTG formatted data will be output to
+                     a single activity being performed repeatedly. The parameter is a string of the name of the activity
+    :param args.ptg_out_file: The absolute path of the feather file will will output that will contain the PTG data format.
```
Purg, line 20: Duplicate use of "will". Maybe the first one was supposed to be "we"?
```suggestion
    :param args.ptg_out_file: The absolute path of the feather file will output that will contain the PTG data format.
```

scripts/parse_footpedal_anns.py
```diff
@@ -0,0 +1,118 @@
...
+
+    imgs = sorted(os.listdir(image_path))
+    event_data = []
+    with open(ann_path, "r") as inFile:
+        event_data = ast.literal_eval(inFile.read())
+
+    events = {}
+    num_anns = 0
+    num_errors = 0
+    #Extract the combined timestamp from the annotation_event_data.txt file
```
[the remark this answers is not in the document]
Purg: By "constraints" I meant to refer to the "Start annotation", "Start error", etc. constant values. Looking at this script in isolation they seem "out of no where", so leaving a bread-crumb to where these come from is good to document.

scripts/parse_footpedal_anns.py
```diff
@@ -101,8 +103,8 @@ def main(args):
 
                 track_json['tracks'][detection] = track_dict
         json_obj = json.dumps(track_json)
```
Purg, line 105: We can remove this line with the change to using `json.dump` below.

scripts/parse_footpedal_anns.py
```diff
@@ -0,0 +1,123 @@
...
+    events = {}
+    num_anns = 0
+    num_errors = 0
+
+    imgs = sorted(os.listdir(args.image_path))
+    event_data = []
+    with open(args.ann_path, "r") as inFile:
+        event_data = json.loads(inFile.read())
+
+    # Extract the combined timestamp from the annotation_event_data.txt file and count number of annoataions and errors
```
Purg, line 36:
```suggestion
    # Extract the combined timestamp from the annotation_event_data.json file and count number of annoataions and errors
```

Purg (approved):

Purg (approved):

PR #104, opened 2022-08-17
Add depth camera topics to ROS bag recording

joshanderson-kw:


joshanderson-kw:
@alexander-lynch, with this and #103, you should be able to record bags with the mono16 depth images. I still need to work on the bag exploder script, but hopefully this will let you at least record all of your data.

Purg (approved):

PR #110, opened 2022-09-19
Add ROC curve

hdefazio:


Purg:
Rebasing on current master.

angel_system/eval/visualization.py
```diff
@@ -151,3 +155,78 @@ def plot_pr_curve(self, detect_intersection_thr=0.1):
...
+
+            fpr[i], tpr[i], _ = roc_curve(truth, pred)
+            roc_auc[i] = auc(fpr[i], tpr[i])
+
+            ax.plot(
+                fpr[i],
+                tpr[i],
+                color=colors[i],
+                lw=2,
+                label="ROC curve of class {0} (area = {1:0.2f})".format(i, roc_auc[i]),
```
Purg, line 196: I just noticed this was the case for the PR plot output as well, but is it possible/sensible to label the semantic class name here instead of just an integer? This is edging towards bikeshedding, which I don't want to enter, but I think including the semantic label here will help with understanding what the plot is showing without going to external resources (e.g. needing to maintain a specific label file with plots being output, putting plots into slides, etc.).

angel_system/eval/visualization.py
```diff
@@ -151,3 +155,78 @@ def plot_pr_curve(self, detect_intersection_thr=0.1):
...
+
+        colors = plt.cm.rainbow(np.linspace(0, 1, len(self.labels)))
+
+        # ============================
+        # Get ROC and AUC per class 
+        # ============================
+        fpr = dict()
+        tpr = dict()
+        roc_auc = dict()
+        for i, row in self.labels.iterrows():
```
Purg, line 179: Also intersecting the PR curve output a little here: This global for-loop is adding all labels to a single plot. This will lead to overcrowding as we get more and more class labels into the plot. We may want to have an option to, for both PR and ROC curves, output single-class plots out to independent files, or express a limit to the number of curves on a single plot. E.g. if there are 30 labels, limit only 5 at a time being rendered on a plot, thus creating 6 files.

hdefazio: I added subfolders for roc and pr and switched it so it writes out individual label plots in addition to the overall plot.

PR #112, opened 2022-09-29
Add UHO activity detector

joshanderson-kw:
Adds Unified Hand Object (UHO) model detector node


joshanderson-kw:
@Purg, FYI, in case you want to start reviewing.

joshanderson-kw:
@Purg, this publishes the `ActivityDetection` message now, so should be ready to review!

ros/angel_system_nodes/angel_system_nodes/uho_activity_detector.py
```diff
@@ -0,0 +1,341 @@
+# TODO: This is added so that the the angel_system/uho/src folder
+# is found when torch attempts to load the saved model checkpoints.
+# Is there a better way to do this?
+import sys
+sys.path.append("angel_system/uho")
```
Purg, line 5: This seems to only be required because of the additional content in the epoch checkpoint outside of the state dict.

ros/angel_system_nodes/angel_system_nodes/uho_activity_detector.py
```diff
@@ -0,0 +1,341 @@
...
+        lhand, rhand = self.get_hand_pose_from_msg(hand_pose)
+        self._hand_poses['lhand'].append(lhand)
+        self._hand_poses['rhand'].append(rhand)
+
+    def obj_det_callback(self, msg):
+        """
+        Callback for the object detection set message. Once there is a object
+        detection message for each frame in the current set, the activity detector
+        model is called and a new activity detection message is published with
+        the current activity predictions.
```
Purg, line 165: I think this description is no longer accurate: this method is no longer associating detections to every frame.

ros/angel_system_nodes/angel_system_nodes/uho_activity_detector.py
```diff
@@ -0,0 +1,341 @@
...
+                        # so keep it for later
+                        continue
+
+                det_descriptors = (
+                    torch.Tensor(det.descriptors).reshape((det.num_detections, det.descriptor_dim))
+                )
+                aux_data['dets'].append(det_descriptors[:self._topk])
+
+                bboxes = [
+                    torch.Tensor((det.left[i], det.top[i], det.right[i], det.bottom[i])) for i in range(self._topk)
```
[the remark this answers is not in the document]
joshanderson-kw: Oh, good catch! Yeah, I think the original way was not sending the actual topk detections, just the first 5. I added some logic to sort the detections first.

ros/angel_system_nodes/angel_system_nodes/uho_activity_detector.py
```diff
@@ -0,0 +1,341 @@
...
+        log.info(f"Labels: {self._labels_file}")
+
+        self.subscription_list = []
+        # Image subscription
+        self.subscription_list.append(mf.Subscriber(self, Image, self._image_topic))
+        # Hand pose subscription
+        self.subscription_list.append(mf.Subscriber(self, HandJointPosesUpdate, self._hand_topic))
+        self.time_sync = mf.ApproximateTimeSynchronizer(
+            self.subscription_list,
+            1500, # TODO: queue size
```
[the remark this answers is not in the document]
joshanderson-kw: Yeah, this was leftover from some debugging. It definitely could be smaller.

There's a good bit of lag, at least a few seconds. I adjusted the values a bit but I'm sure there is more fine-tuning that we could do.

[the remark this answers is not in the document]
Purg: I definitely don't want to bikeshed. If you don't have some more constrained values in mind, then maybe leave them as is until we have a more informed judgement.

ros/angel_system_nodes/angel_system_nodes/uho_activity_detector.py
```diff
@@ -0,0 +1,336 @@
...
+                                      m.pose.position.y,
+                                      m.pose.position.z]}
+                      for m in msg.joints]
+
+        # Rejecting joints not in OpenPose hand skeleton format
+        reject_joint_list = {'ThumbMetacarpalJoint',
+                            'IndexMetacarpal',
+                            'MiddleMetacarpal',
+                            'RingMetacarpal',
+                            'PinkyMetacarpal'}
```
Purg, line 300: Nitpick: white-spacing doesn't align values.

ros/angel_system_nodes/angel_system_nodes/uho_activity_detector.py
```diff
@@ -0,0 +1,336 @@
...
+
+                # Get the topk detection confidences
+                det_confidences = (
+                    torch.Tensor(det.label_confidences)
+                    .reshape((det.num_detections, len(det.label_vec)))
+                )
+                topk, i = torch.topk(det_confidences.flatten(), self._topk)
+
+                ind_2d = np.array(np.unravel_index(i.numpy(), det_confidences.shape)).T
+                top_det_idx = ind_2d[:, 0]
```
Purg, line 230: I'm not sure if this is what we want. Since this is flattening *all* confidences, it can be the case where the top confidences are in the same descriptor (all descriptors "even", one has two spikes).

The following looks as descriptor max confidence values specifically before going into the topk function, should avoid jumping out of torch format (maybe the tiniest amount more efficient), and more readable IMO.
```suggestion
                det_max_confidences = det_confidences.max(axis=1).values
                _, top_det_idx = torch.topk(det_max_confidences, self._topk)
```

Purg (approved):

PR #116, opened 2022-10-10
First-pass centralized activity label configuration

Purg:
The intent of this config file is to centralize the ID value, the label strings, and the step description across actors that want to make use of this information.

### DIFF
--- a/config/activity_labels/recipe_coffee.yaml
+++ b/config/activity_labels/recipe_coffee.yaml
@@ -1,0 +1,60 @@
+version: "1"
+labels:
+  # Item:
+  # - id: Integer identification number of the label.
+  # - label: Short human-readable semantic string to succinctly represent the
+  #          label.
+  - id: 0
+    label: "background"
+  - id: 1
+    label: "measure-12oz-water"
+  - id: 2
+    label: "pour-water-kettle"
+  - id: 3
+    label: "turn-kettle-on"
+  - id: 4
+    label: "place-dipper-on-mug"
+  - id: 5
+    label: "filter-fold-half"
+  - id: 6
+    label: "filter-fold-quarter"
+  - id: 7
+    label: "place-filter"
+  - id: 8
+    label: "spread-filter"
+  - id: 9
+    label: "scale-turn-on"
+  - id: 10
+    label: "place-bowl-on-scale"
+  - id: 11
+    label: "zero-scale"
+  - id: 12
+    label: "measure-coffee-beans"
+  - id: 13
+    label: "pour-coffee-grinder"
+  - id: 14
+    label: "set-timer-20"
+  - id: 15
+    label: "timer-turn-on"
+  - id: 16
+    label: "grind-beans"
+  - id: 17
+    label: "pour-beans-filter"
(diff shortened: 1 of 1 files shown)


Purg:
@hdefazio @mattbrown11

config/activity_labels/recipe_coffee.yaml
```diff
@@ -0,0 +1,101 @@
+version: "1"
+labels:
+  # Item:
+  # - id: Integer identification number of the label.
+  # - label: Short human-readable semantic string to succinctly represent the
+  #          label.
+  # - step_detail: Longer descriptive string detailing the labeled step. This
+  #                currently acts as the step "instruction" but, it's presence
```
mattbrown11, line 8: "instruction," but :)

config/activity_labels/recipe_coffee.yaml
```diff
@@ -0,0 +1,101 @@
...
+      Pour the water from the liquid measuring cup into the electric
+      kettle
+  - id: 3
+    label: "turn-kettle-on"
+    step_detail: "Turn on the Kettle"
+  - id: 4
+    label: "place-dipper-on-mug"
+    step_detail: "Place the dripper on top of the mug"
+  - id: 5
+    label: "filter-fold-1"
```
mattbrown11, line 28: filter-fold-half

config/activity_labels/recipe_coffee.yaml
```diff
@@ -0,0 +1,101 @@
...
+    step_detail: >-
+      Take the coffee filter and fold it in half to create a semi-circle
+  - id: 6
+    label: "filter-fold-2"
```
mattbrown11, line 32: filter-fold-quarter
fold-1 and fold-2 sort of conveys something about order as if "this is the folding that happens first" and then "this is the folding that happens second." The activities need to be identifiable given only the content of the short video segment.

config/activity_labels/recipe_coffee.yaml
```diff
@@ -0,0 +1,101 @@
...
+    step_detail: "Turn on the timer"
+  - id: 16
+    label: "grind-beans"
+    step_detail: >-
+      Grind the coffee beans by pressing and holding down on the black part of
+      the lid
+  - id: 17
+    label: "pour-beans-filter"
+    step_detail: >-
+      Pour the grounded coffee beans into the filter cone prepared in step 2
```
mattbrown11, line 71: Remove reference to step 2

config/activity_labels/recipe_coffee.yaml
Purg: Sure. The previous choice was arbitrary so I'll go with your suggestion here.

config/activity_labels/recipe_coffee.yaml
Purg: This was a raw copy-paste of the sentences from the steps doc, but I believe this will be moot as I will be removing the step detail field here as mentioned in chat.

Purg:
Removed the step details to be left for step-related configurations.

PR #117, opened 2022-10-11
Remove time synchronizer

joshanderson-kw:
Remove ROS time synchronizer use
Move/reformat activity detector node helper functions to angel_utils & angel_system


joshanderson-kw:
@Purg, please take a look if you have time. I think this is what we talked about with regards to refactoring the activity detector node.

Purg:
This is OK for now. Chunks of this will likely be revised in my next task.

PR #118, opened 2022-10-11
New task steps config

Purg:
Revision to the task steps configuration format for the new HMM direction for a future revised task monitor node. This is intended to replace the `mit_ll_eval_one_coffee_recipe_steps.json` when the time comes.

### DIFF
--- a/config/tasks/task_steps_config-recipe_coffee.yaml
+++ b/config/tasks/task_steps_config-recipe_coffee.yaml
@@ -12,116 +12,138 @@ steps:
   #                step. This must reference an ID within the `activity_labels`
   #                configuration file referenced above.
   # - description: Human semantic description of this step.
-  # - median_duration_seconds: [Optional] Median expected time this task will=
+  # - median_duration_seconds: Median expected time this task will
   #                            consume in seconds.
   - id: 0
     activity_id: 1
     description: >-
       Pour 12 ounces of water into liquid measuring cup
-    median_duration_seconds: 10
+    median_duration_seconds: 5
   - id: 1
     activity_id: 2
     description: >-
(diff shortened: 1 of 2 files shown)


Purg:
@alexander-lynch Matt suggested that you complete this PR, filling in the part marked "TODO".

mattbrown11:
LGTM

Purg:
I'm adding a couple comments to describe fields in the config.

config/tasks/task_steps_config-recipe_coffee.yaml
```diff
@@ -0,0 +1,117 @@
...
+    activity_id: 25
+    description: >-
+      Remove the coffee grounds and paper filter from the dripper
+  - id: 25
+    activity_id: 26
+    description: >-
+      Discard the coffee grounds and paper filter
+
+hmm:
+  num_steps_can_jump_fwd: 1
```
mattbrown11, line 150: Can we add
dt : Time (seconds) between time steps of HMM.

mattbrown11:
We don't have a reference to a background class. We have a few options here.
1) We can enforce that the first step of every recipe is 'background', and then my HMM class can parse this yaml and infer that the first step (and the associated activity class enum) is a background activity. This way, my HMM knows which activity enum to treat as a background action.
2) We explicitly do not include a background step in the recipe. But internally, my HMM needs to assign a background action, so it needs to know where to find the confidence for that action in the classifier confidences. Background could always be index 0 or always called 'background' if we are going with string-based encoding.

Purg:
> We don't have a reference to a background class. ...

Your HMM parameterization could have a parameter like `activity_background_class_id` that we set to an integer, e.g. `0`.

PR #119, opened 2022-10-11
Demo UI

hdefazio:
New: 
- Demo UI
- recipe configs for the new model

Updated: 
- Added log statement to detector
- Added recipe name to QueryTaskGraph


hdefazio:
@Purg PTAL

(this was also tested by @alexander-lynch on the live system)

ros/angel_utils/demo_ui/main.html
```diff
@@ -0,0 +1,81 @@
...
+  <div id="bottom-row">
+    <div id="activity">
+      <div class="highlighted-title">
+        <span class="icon-hex"></span>
+        <span class="text title-hex">Activity Classification</span>
+      </div>
+      
+    <canvas id="activity-conf"></canvas>
+    <script type="text/javascript" src="src/activity_conf.js"></script>
+    <script type="text/javascript" src="src/task_list.js"></script>
```
Purg, line 48: Not inline with the current indentation level: increase indentation one level.

docker/workspace-build/Dockerfile
```diff
@@ -38,6 +40,9 @@ RUN apt-get -y update \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/*
 
+# For the Demo UI
+RUN source /opt/ros/${ROS_DISTRO}/setup.bash
+
```
Purg, line 45: This has no effect beyond this layer. Is this "cruft" at this point, or was there an intention behind keeping this around?

ros/angel_utils/demo_ui/src/task_list.js
```diff
@@ -0,0 +1,108 @@
...
+  var next_task_name = m.current_step;
+  var next_checkbox = document.getElementById(next_task_name);
+
+  if(next_checkbox != null){
+    var next_existing_check = next_checkbox.querySelector("#checkmark");
+
+    if(next_existing_check != null){
+      next_checkbox.removeChild(next_existing_check);
+    }
+  }
```
Purg, line 82: This is fine for going back one, but I interpret that if we went back to the beginning then would this only remove one checkmark out of many?

ros/angel_utils/demo_ui/src/task_list.js
```diff
@@ -0,0 +1,108 @@
...
+
+  // Update colors in chart
+  var chart = Chart.getChart('activity-conf');
+  var colors = new Array(xValues.length).fill("rgba(0, 104, 199, 1.0)");
```
Purg, line 86: Where does `xValues` come from? Is this supposed to be supplied by `chart`?

tmux/demo-ui.yml
```diff
@@ -0,0 +1,91 @@
...
+      -r __ns:=${ROS_NAMESPACE}
+      -p topic_input_images:=PVFramesRGB
+      -p topic_input_det_2d:=ObjectDetections
+      -p topic_output_images:=pv_image_detections_2d
+  - compress_image: ros2 run image_transport republish raw compressed --ros-args 
+       -r __ns:=${ROS_NAMESPACE}
+       --remap in:=pv_image_detections_2d 
+       --remap out/compressed:=pv_image_detections_2d/compressed
+  # Demo UI
+  - demo_ui: ros2 launch src/angel_utils/demo_ui/launch/websocket.launch
```
Purg, line 91: Could this be told to also operate under the same `ROS_NAMESPACE` as everything else?

ros/angel_system_nodes/angel_system_nodes/task_monitor.py
```diff
@@ -294,6 +294,7 @@ def query_task_graph_callback(self, request, response):
         task_g.task_steps = self._task.task_graph_steps
         task_g.task_levels = self._task.task_graph_step_levels
 
+        response.task_title = self._task.name
```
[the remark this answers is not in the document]
joshanderson-kw: Yeah, I need to fill this in for the new node. I'll do that soon.

ros/angel_system_nodes/configs/mit_ll_eval_one_coffee_recipe_steps_v2.json
```diff
@@ -0,0 +1,187 @@
+{
+  "Measure 12 ounces of cold water and transfer to a kettle. Boil the water.": {
```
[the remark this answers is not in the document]
hdefazio: After pulling latest (with HMM) - are these files even used?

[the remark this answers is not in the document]
Purg: I think the only reason we would have any version of this format is to support an example of what the V1 task monitor currently takes in (it's not updated to take in the new format as of yet). I would say keep the first version of these files around, but if you're not using the V2 versions you added here, remove them.

Purg (approved):

PR #120, opened 2022-10-12
Create angel_system UHO activity detector api

joshanderson-kw:
Adds `AuxData` dataclass, and `predict` and `get_model` functions

### DIFF
--- a/angel_system/uho/src/data_helper.py
+++ b/angel_system/uho/src/data_helper.py
@@ -1,13 +1,38 @@
 """
 Helper functions for forming the input data to the UHO model.
 """
+from dataclasses import dataclass
 from typing import Callable
 from typing import Dict
 from typing import List
+from typing import Tuple
 
 from angel_msgs.msg import ObjectDetection2dSet
 import numpy as np
 import torch
+
+
+@dataclass
+class AuxData:
+    """
+    Class for representing the aux data dictionary required by the UHO module.
+
+    Attributes:
+        lhand: List of np.ndarrays of size (63,) representing the 63 joint poses
+            of the left hand.
+        rhand: List of np.ndarrays of size (63,) representing the 63 joint poses
+            of the right hand.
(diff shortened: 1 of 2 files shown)


angel_system/uho/src/predict.py
```diff
@@ -0,0 +1,68 @@
...
+    )
+
+    detector: UnifiedHOModule = UnifiedHOModule(
+        fcn=fcn,
+        temporal=temporal,
+        checkpoint=checkpoint_path,
+        device=device,
+        labels_file=labels_path
+    )
+    detector.eval()
```
Purg, line 54: May want to make a note in the doc-string that the model is for evaluation only given the presence of this here.

angel_system/uho/src/predict.py
```diff
@@ -0,0 +1,68 @@
...
+    return detector.to(device=device)
+
+
+def predict(
+    model: UnifiedHOModule,
+    frames: List[np.ndarray],
+    aux_data: AuxData
+) -> Tuple[Tuple[torch.Tensor, torch.Tensor], List[str]]:
+    """
+    Returns the result of the forward call for the UHO model.
```
Purg, line 64: We need to add an explanation as to the format that is being output here.

angel_system/uho/src/data_helper.py
```diff
@@ -1,26 +1,51 @@
...
 
 from angel_msgs.msg import ObjectDetection2dSet
 import numpy as np
 import torch
 
+
+@dataclass
+class AuxData:
+    """
+    Class for representing the aux data dictionary required by the UHO module.
```
Purg, line 18: We could add a note that it is expected that all list attributes are expected to be the same length.

Purg (commented):
Pop the `predict.py` module out to the `angel_system/uho/` level.

Purg (approved):

PR #128, opened 2022-10-17
Dev/add hmm node

joshanderson-kw:
Add ROS node for HMM class from angel_system.

TODO: Not all of the `TaskUpdate` fields are filled in yet (name, description, current activity, next activity, time remaining), but not sure how if those are even used.


joshanderson-kw:
@mattbrown11, @Purg, please take a look.

angel_system/activity_hmm/core.py
```diff
@@ -236,7 +236,7 @@ def __init__(self, dt, class_str, med_class_duration,
         class_str_.append('background')
         class_str_map.append(len(class_str_) - 1)
         bckg_mask.append(True)
-        inv_map.append(i)
+        inv_map.append(0)
```
mattbrown11, line 239: Good you fixed this. I made another PR for this, but I'll just let it be fixed here.

mattbrown11:
LGTM

angel_system/activity_hmm/core.py
Purg: Also in your other branch, so we'll see who gets in first and then the other will likely need to rebase. :)

ros/angel_system_nodes/setup.py
```diff
@@ -33,6 +33,7 @@
                 'annotation_event_monitor = angel_system_nodes.annotation_event_monitor:main',
                 'object_detector_with_descriptors = angel_system_nodes.object_detector_with_descriptors:main',
                 'uho_activity_detector = angel_system_nodes.uho_activity_detector:main',
+                'hmm = angel_system_nodes.hmm:main',
```
Purg, line 36: This name is confusing and lacks context. I would more recommend that we rename the existing task monitor to something like `task_monitor_v1` and this `task_monitor_v2` as they fulfill the same role (to my understanding). Of course, reflect that name change to the V1 node in the existing config files to maintain relevance.

It would be good to start a .rst file under `docs/task_monitor.rst` that lists the relevant nodes and a short description about what the difference is (e.g. v1 is the first iteration that combined activity classifications and configuration to very simply make task step determinations, and v2 uses a hidden markov model to deduce step transitions). This initial doc doesn't need to be more detailed than that, we just need to begin tracing history before we lose track of it.

ros/angel_system_nodes/angel_system_nodes/hmm.py
```diff
@@ -0,0 +1,254 @@
...
+            .string_value
+        )
+        self._task_state_topic = (
+            self.declare_parameter("task_state_topic", "TaskUpdates")
+            .get_parameter_value()
+            .string_value
+        )
+
+        # Instantiate the HMM module
+        self._hmm = ActivityHMMRos(self._config_file)
```
Purg, line 44: I see that this is being used across different threads. I don't think I saw any thread safety in this object, so you likely need to create an associated lock to guard the usage of this resource throughout this node.

ros/angel_system_nodes/angel_system_nodes/hmm.py
```diff
@@ -0,0 +1,254 @@
...
+                # Get the HMM prediction
+                start_time = time.time()
+                self._current_step = self._hmm.get_current_state()
+                log.info(f"HMM get_current_state time: {time.time() - start_time}")
+                log.info(f"Current step {self._current_step}")
+
+                # Publish a new TaskUpdate message
+                self.publish_task_state_message()
+
+    def thread_run_hmm_skip(self):
```
Purg, line 190: Does this need to be a separate thread? If so, the rationale should be included in the description of the method.

ros/angel_system_nodes/angel_system_nodes/hmm.py
```diff
@@ -0,0 +1,254 @@
...
+        """
+        Check that the HMM runtime is still alive and raise an exception
+        if it is not.
+        """
+        alive = (
+            self._hmm_state_thread.is_alive() and self._hmm_skip_thread.is_alive()
+        )
+        if not alive:
+            self.get_logger().warn("HMM thread no longer alive.")
+            self._hmm_thread.join()
```
Purg, line 222: If you do want to stick with both runtime threads, should both be stopped if one stops? If so, then this conditional should likely clear both event variables and then join on both threads before continuing.

ros/angel_system_nodes/angel_system_nodes/hmm.py
```diff
@@ -0,0 +1,254 @@
...
+                activity_msg.conf_vec,
+                source_stamp_start_frame_sec,
+                source_stamp_end_frame_sec,
+            )
+
+            # Tell the HMM threads to wake up
+            self._hmm_state_awake_evt.set()
+            self._hmm_skip_awake_evt.set()
+
+    def publish_task_state_message(self, activity=None):
```
Purg, line 119: The parameter isn't used / documented. Remove?

Purg (commented):
Node state locking is a little lacking, but may be OK enough by context... If the HMM prediction is quick, threading this may be complexity overkill.

ros/angel_system_nodes/angel_system_nodes/hmm.py
```diff
@@ -0,0 +1,254 @@
...
+            message.previous_step = self._previous_step
+
+        # TODO: Populate task name and description
+        # TODO: Do we need to fill in current/next activity?
+        # TODO: Fill in time remaining?
+
+        self._publisher.publish(message)
+
+        # Update state tracking vars
+        self._previous_step = self._current_step
```
Purg, line 149: The previous and current step may be the same a lot of the time given the current stream prediction context that I'm seeing here. Is that OK? IIRC this was supposed to be the "true" previous step in the task regardless of how long we have been on the current step?

ros/angel_system_nodes/angel_system_nodes/task_monitor_v2.py
```diff
@@ -0,0 +1,282 @@
...
+
+    def query_task_graph_callback(self, request, response):
+        """
+        Populate the `QueryTaskGraph` response with the task list
+        and return it.
+        """
+        log = self.get_logger()
+        task_g = TaskGraph()
+
+        task_g.task_steps = self._hmm.model.class_str
```
Purg, line 169: Missing hmm lock context.

ros/angel_system_nodes/angel_system_nodes/task_monitor_v2.py
```diff
@@ -0,0 +1,282 @@
...
+        message = TaskUpdate()
+
+        # Populate message header
+        message.header.stamp = self.get_clock().now().to_msg()
+        message.header.frame_id = "Task message"
+
+        # Populate steps and current step
+        message.steps = self._hmm.model.class_str
+        message.current_step = self._current_step
+        message.current_step_id = self._hmm.model.class_str.index(self._current_step)
```
Purg, line 146: Missing hmm lock context.

ros/angel_system_nodes/angel_system_nodes/task_monitor_v2.py
```diff
@@ -0,0 +1,282 @@
...
+        """
+        Check that the HMM runtime is still alive and raise an exception
+        if it is not.
+        """
+        alive = (
+            self._hmm_state_thread.is_alive() and self._hmm_skip_thread.is_alive()
+        )
+        if not alive:
+            self.get_logger().warn("HMM thread no longer alive.")
+            self._hmm_active.clear()  # make HMM active flag "False"
```
Purg, line 248: both events need to be cleared because either may have triggered entering this block.

Purg (approved):

PR #131, opened 2022-10-18
Fix bug in populating hmm mean and covariance.

mattbrown11:
Fixed a bug in how the HMM's mean and covariance are populated in the  ActivityHMM.__init__, which yields much better results. Added a `load_and_discretize_data` function to efficiently read group truth and detection json.

### DIFF
--- a/angel_system/activity_hmm/core.py
+++ b/angel_system/activity_hmm/core.py
@@ -4,6 +4,10 @@ from sklearn.metrics import precision_recall_curve
 import pandas as pd
 import json
 import yaml
+import re
+from ast import literal_eval
+
+from angel_system.eval.support_functions import time_from_name
 
 try:
     import matplotlib.pyplot as plt
@@ -236,7 +240,7 @@ class ActivityHMM(object):
         class_str_.append('background')
         class_str_map.append(len(class_str_) - 1)
         bckg_mask.append(True)
-        inv_map.append(i)
+        inv_map.append(0)
 
         # Initialize the remainder of the classes
         for i, clss in enumerate(class_str[1:], start=1):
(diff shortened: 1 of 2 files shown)


mattbrown11:
@Purg @joshanderson-kw PTAL. simulate_data.py is my sandboxy script, so I wouldn't spend too much effort reviewing it.

joshanderson-kw (approved):
LGTM, I didn't look at the simulate_data script or the helper functions though.

angel_system/activity_hmm/core.py
```diff
@@ -743,11 +773,219 @@ def score_raw_detections(X, Z, plot_results=False):
...
+            'start': time_from_name(row["start_frame"]),
+            #'end': float(RE_FILENAME_TIME.match(row["end_frame"]).groups()[0])
+            'end': time_from_name(row["end_frame"])
+        }
+        gt.append(g)
+
+    print(f"Loaded ground truth from {activity_gt}")
+    gt = pd.DataFrame(gt)
+
+    detections_input = [det for det in (literal_eval(s) for s in open(extracted_activity_detections))][0]
```
Purg, line 859: `json.load` please, not AST. If the source file is not formatted correctly for `json.load` to operate, that is a bug and it should be fixed.

angel_system/activity_hmm/core.py
```diff
@@ -275,37 +279,63 @@ def __init__(self, dt, class_str, med_class_duration,
...
 
-            class_conf_mean_mat = np.zeros((N_, N_))
-            sphere_conf_cov_mat = np.zeros(N_)
-
-            # Square to turn from std to cov. We'll reuse the name.
-            class_std_conf = class_std_conf**2
+            if np.ndim(class_mean_conf) == 1:
+                class_mean_conf = np.diag(class_mean_conf)
+            elif np.ndim(class_mean_conf) > 2:
+                raise Exception('np.ndim(class_mean_conf) must be 1 or 2')
```
Purg, line 300: Generally do not use base-class exceptions, [concrete exceptions](https://docs.python.org/3/library/exceptions.html#concrete-exceptions) should be used if looking for a builtin, define your own exception, or use an assert.

angel_system/activity_hmm/core.py
```diff
@@ -743,11 +773,219 @@ def score_raw_detections(X, Z, plot_results=False):
...
+
     return auc
+
+
+def load_and_discretize_data(activity_gt: str,
+                             extracted_activity_detections: str,
+                             time_window: float, uncertain_pad: float):
+    """
+
+    Parameters
```
Purg, line 819: Missing description about what this is supposed to do.

angel_system/activity_hmm/core.py
```diff
@@ -743,11 +773,219 @@ def score_raw_detections(X, Z, plot_results=False):
...
+    -------
+    TYPE
+        DESCRIPTION.
+
+    """
+    gt_f = pd.read_feather(activity_gt)
+    # Keys: class, start_frame,  end_frame, exploded_ros_bag_path
+
+    gt = []
+    RE_FILENAME_TIME = re.compile(r"frame_\d+_(\d+)_\d+.\w+")
```
Purg, line 845: This is no longer used, I only see it in commented-out code, which could also be removed.

mattbrown11: Addressed.

angel_system/activity_hmm/core.py
mattbrown11: Addressed.

angel_system/activity_hmm/core.py
mattbrown11: Addressed.

angel_system/activity_hmm/core.py
mattbrown11: Addressed.

Purg (approved):

PR #132, opened 2022-10-18
Add stub activity classifier mode to UHO detector node

joshanderson-kw:
Add `gt_file` param to the UHO activity detector node. If no file is provided, the node operates as usual running the activity classifier. If a a ground truth file is provided, the node uses `gt_predict` instead of calling the classifier predict function.

Tested on one Alex's recordings and it seems to match up pretty well.

### DIFF
--- a/angel_system/eval/gt_predict.py
+++ b/angel_system/eval/gt_predict.py
@@ -1,0 +1,80 @@
+from typing import (
+    List,
+    Tuple,
+)
+
+import pandas
+import torch
+
+from angel_system.eval.support_functions import time_from_name
+
+
+# TODO: What is the best place for this function?
+def gt_predict(
+    gt_file: str,
+    frame_set_start_stamp: float,
+    frame_set_end_stamp: float
+) -> Tuple[Tuple[torch.Tensor, torch.Tensor], List[str]]:
+    """
+    Creates a simulated classifier prediction given the ground truth annotation
+    file and the frame timestamps. The frame stamps are used to determine which
+    activity was being performed in the ground truth file.
+
+    The return type is consistent with the UHO's predict function.
+
+    :param gt_file: string path for the ground truth annotations to use. It is
+        expected that this is a feather file.
+    :param frame_set_start_stamp: The start time in seconds for this frame
+        window.
+    :param frame_set_end_stamp: The end time in seconds for this frame window.
+
+    :return: A tuple consisting of two elements:
+            0: Classifier results: A tuple consisting of a [1 x n_classes] tensor
+               representing the classifier's confidence for each class and a tensor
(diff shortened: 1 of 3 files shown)


angel_system/eval/gt_predict.py
```diff
@@ -0,0 +1,87 @@
...
+    Tuple,
+)
+
+import pandas
+import torch
+
+from angel_system.eval.support_functions import time_from_name
+
+
+# TODO: What is the best place for this function?
```
Purg, line 13: I think this could be in `angel_system/utils/activity_classification.py` instead of here. I happen to interpret that this function is not assisting the evaluation code/tool in specific.

ros/angel_system_nodes/angel_system_nodes/uho_activity_detector.py
```diff
@@ -565,6 +601,11 @@ def thread_predict_runtime(self):
             self._window_criterion_new_leading_frame,
         ]
 
+        if self._overlapping_mode == False:
```
Purg, line 604:
```suggestion
        if self._overlapping_mode is False:
```

ros/angel_system_nodes/angel_system_nodes/uho_activity_detector.py
```diff
@@ -663,7 +704,16 @@ def _process_window(self, window: InputWindow) -> ActivityDetection:
...
+
+            if self._gt_file == "":
+                # No ground truth provided, use the detector
+                pred_conf, pred_labels = predict(self._detector, frame_set, aux_data)
+            else:
+                pred_conf, pred_labels = gt_predict(
+                    self._gt_file,
+                    time_to_int(window.frames[0][0]) * 1e-9,
+                    time_to_int(window.frames[-1][0]) * 1e-9,
+                )
```
Purg, line 716: This isn't what I had in mind for swapping the "prediction" function. I meant that we could swap out the "window processing" function in the run-time loop. Before the loop we would assign to, say, `process_window_fn` either this method, `self._process_window`, or a ground-truth version, e.g. `self._process_window_gt`. The ground truth version of processing doesn't care about the pre- and post-processing steps that occur here. We also want to separate responsibilities for good maintainability practice for when we inevitably need to come back in here and modify the true-prediction side of things.

angel_system/eval/gt_predict.py
joshanderson-kw: Done

ros/angel_system_nodes/angel_system_nodes/uho_activity_detector.py
joshanderson-kw: Done

Purg (approved):

PR #137, opened 2022-10-24
Add initial setup for Sphinx documentation space support

Purg:
Low priority, but added Sphinx documentation support to build off of.

* Added Sphinx development deps
* Update `.gitignore` to include python template


.gitignore
```diff
@@ -6,13 +6,154 @@
...
+*.py,cover
+.hypothesis/
+.pytest_cache/
+cover/
+
+# Translations
+*.mo
+*.pot
+
+# Django stuff:
```
joshanderson-kw, line 67: Do we need the django/celery stuff here?

joshanderson-kw (approved):

.gitignore
Purg: Responded offline: The content of the block was populated via a template, so there is more than just the most basic python stuff in that. We will likely not hit all the ignores that are populated here, but they will also not harm anything unless there is a conflict in the future, but we can mitigate that then as the issue will be quite explicit.

PR #138, opened 2022-10-24
Add keyboard skip to task_monitor_v2

joshanderson-kw:
Add next/previous step shortcuts
Remove publishing of background step

### DIFF
--- a/ros/angel_system_nodes/angel_system_nodes/task_monitor_v2.py
+++ b/ros/angel_system_nodes/angel_system_nodes/task_monitor_v2.py
@@ -6,6 +6,7 @@ from threading import (
 import time
 import yaml
 
+from pynput import keyboard
 import rclpy
 from rclpy.node import Node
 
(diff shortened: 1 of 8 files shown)


ros/angel_system_nodes/angel_system_nodes/task_monitor_v2.py
```diff
@@ -168,10 +188,22 @@ def publish_task_state_message(self):
...
+            else:
+                hmm_current_step_id = message.steps.index(
+                    self._current_step
+                )
+                message.current_step_id = hmm_current_step_id + 1
+
+        if message.current_step_id == len(message.steps):
+            # Because we don't have a done step, keep this previous step but
+            # advance the index so the ARUI can see we're done
+            message.current_step = message.steps[message.current_step_id - 1]
```
Purg, line 204: Is this describing the mechanic in which we signify the task is "done"? If so, I think it would be more understandable to be more explicit about this and use the `task_complete_confidence` field we added (set to `1.0` here).

ros/angel_system_nodes/angel_system_nodes/task_monitor_v2.py
```diff
@@ -250,15 +282,18 @@ def thread_run_hmm(self):
                     self._skip_score = self._hmm.get_skip_score()
                 log.info(f"HMM computation time: {time.time() - start_time}")
 
-                if self._current_step != step:
+                step_id = self._hmm.model.class_str.index(step)
+
+                # Only change steps if we have a new step and it is not background
+                if self._current_step != step and step_id != 0:
                     self._previous_step = self._current_step
                     self._current_step = step
```
Purg, line 290: Is this current formulation, are we reporting 1-indexed steps now? That might be at odds with the natural usage downstream, i.e. we may want to convert to zero-indexed values going into the message we output in order to be congruent with the task graph we output via the service also exposed here.

PR #141, opened 2022-10-25
Add revert_to_step method.

mattbrown11:
@Purg @joshanderson-kw

### DIFF
--- a/angel_system/activity_hmm/core.py
+++ b/angel_system/activity_hmm/core.py
@@ -75,7 +75,9 @@ class ActivityHMMRos:
         class_mean_conf = np.array(class_mean_conf)
         self.class_mean_conf = class_mean_conf
 
-        self.model = ActivityHMM(self.dt, class_str, med_class_duration,
+        # This is the model that enforces steps are done in order without
+        # skipping steps.
+        self.noskip_model = ActivityHMM(self.dt, class_str, med_class_duration,
                                  num_steps_can_jump_fwd=0,
                                  num_steps_can_jump_bck=0,
                                  class_mean_conf=class_mean_conf,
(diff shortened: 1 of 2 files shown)


angel_system/activity_hmm/core.py
```diff
@@ -178,23 +201,23 @@ def analyze_current_state(self):
             Score indicating how much more likely it is that a step was skipped
             than no steps being skipped.
         """
-        log_prob1, Z_no_skip, X_, Z_no_skip_ = model.decode(X)
-        log_prob1 = model_skip.calc_log_prob_(X_, Z_no_skip_)
+        log_prob1, Z_no_skip, X_, Z_no_skip_ = self.noskip_model.decode(X)
```
Purg, line 204: Use of `X` here is undefined. Did you mean `self.X`?

angel_system/activity_hmm/core.py
```diff
@@ -154,8 +156,29 @@ def add_activity_classification(self, label_vec, conf_vec, start_time,
         self.times = np.hstack([self.times, times_])
         self.X = np.vstack([self.X, X])
 
+    def revert_to_step(self, step_ind):
```
Purg, line 159: Is the "step" index here that of a time-step or the task-step index?

mattbrown11: The integer step as defined in the recipe.yaml via 'id'.

angel_system/activity_hmm/core.py
mattbrown11: Fixed (sorry, playing too fast and loose).

Purg (approved):

PR #147, opened 2022-10-27
New activity class predict

Purg:
This update is somewhat of a hybrid of both @daweidu's updates to the UHO module as well as my integration of it's use into the ROS2 node that makes use of it.


Purg:
@daweidu Could you please take a look at what I've done in the `prediction.py` function `predict`? I would like you to tell me if what I am doing there is sensible or if there is anything that I am doing blatantly wrong. I think discovered some things that could be introspected from the network but I would like your insight of if what I am doing is inherently dangerous or something.

angel_system/uho/prediction.py
```diff
@@ -0,0 +1,335 @@
...
+    """
+    device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+    aux_data = {}
+
+    lhand, rhand = [], []
+    dets, bbox, frms = [], [], []
+    topK = 10
+    for t in batch:
+        lhand = [torch.from_numpy(tmp).to(device) for tmp in t[0]["l_hand"]]
+        rhand = [torch.from_numpy(tmp).to(device) for tmp in t[0]["l_hand"]]
```
alexander-lynch, line 69: These lines look identical, is the second line supposed to be r_hand?

Purg: Probably. I will correct however this function is not actually utilized in prediction (I think Dawei was using it in testing).

PR #149, opened 2022-10-27
Fix import in hmm after eval refactor

hdefazio:


hdefazio:
@Purg / @mattbrown11

Purg:
Already fixing this in the `thursday-fixes` branch as Alex and I QA'd the system this past hour or so.

PR #150, opened 2022-10-27
Thursday fixups

Purg:
Fixes identified through today's QA run-through .

### DIFF
--- a/ros/angel_utils/demo_ui/.gitignore
+++ b/ros/angel_utils/demo_ui/.gitignore
@@ -1,0 +1,1 @@
+/node_modules/
--- a/tmux/2022-11-PI_Eval1-coffee-from_bag_and_truth.yml
+++ b/tmux/2022-11-PI_Eval1-coffee-from_bag_and_truth.yml
@@ -109,4 +109,5 @@ windows:
       -p skip_score_threshold:=300.0
 
   # Currently hard-coded to use the /debug ROS namespace.
-  - engineering_ui: ros2 launch src/angel_utils/demo_ui/launch/websocket.launch
+  - engineering_ui_websocket: ros2 launch rosbridge_server rosbridge_websocket_launch.xml port:=9090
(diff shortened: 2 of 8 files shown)


ros/angel_utils/demo_ui/.gitignore
```diff
@@ -0,0 +1 @@
+/node_modules/
```
hdefazio, line 1: the `npm install` also creates `log`, `build`, and `install` directories that can be ignored

tmux/2022-11-PI_Eval1-coffee-from_bag_and_truth.yml
```diff
@@ -109,4 +109,5 @@ windows:
       -p skip_score_threshold:=300.0
 
   # Currently hard-coded to use the /debug ROS namespace.
-  - engineering_ui: ros2 launch src/angel_utils/demo_ui/launch/websocket.launch
+  - engineering_ui_websocket: ros2 launch rosbridge_server rosbridge_websocket_launch.xml port:=9090
```
hdefazio, line 112: Should `angel_utils/demo_ui` also be renamed?

ros/angel_utils/demo_ui/.gitignore
Purg: Where do these get generated? I don't see them in my space.

tmux/2022-11-PI_Eval1-coffee-from_bag_and_truth.yml
Purg: It could be, but I'm happy to bikeshed that tomorrow at stand-up vs. now.
I was considering that it also maybe doesn't belong in angel_utils, but I also see why it's there, so eh.

ros/angel_utils/demo_ui/.gitignore
hdefazio: hmmm, I see them under demo_ui/ but maybe it's from something else

PR #151, opened 2022-10-28
Dev/test

mattbrown11:
Fixes the issue where `revert_to_step` would place you in background if that was the most-recent state.


mattbrown11:
@Purg @joshanderson-kw

PR #148, opened 2022-10-27
New/updated features in demo-ui

hdefazio:
- [x] Removed done button
- [x] Add task ids to task list
- [x] Added back the colors in activity classification chart
- [x] Add probability of recipe finish to a chart
- [x] Filter displayed bounding boxes to top 5

PR #153, opened 2022-10-28
Add error reporting to demo UI

hdefazio:
Depends on #148


Purg:
Saw this demoed by Alex, and saw the error box pop up. It should be the case that the error message also has a description field that has more info on what step range was skipped over. We should probably include that info in the displayed message.

Purg:
> Saw this demoed by Alex, and saw the error box pop up. It should be the case that the error message also has a description field that has more info on what step range was skipped over. We should probably include that info in the displayed message.

Ha, nevermind, we just realized it could be clicked.

ros/angel_debug/angel_debug/nodes/simple_2d_detection_overlay.cxx
```diff
@@ -296,20 +305,55 @@ ::collect_detections( ObjectDetection2dSet::SharedPtr const det_set )
 
   cv::Point max_point;  // Only x will be populated due to single row scan.
   std::string max_label;
+  double maxVal;
+
+  std::vector<std::string> all_labels;
+  std::vector<double> all_dets;
```
Purg, line 311: Nitpick: This is more an `all_scores` to me.

Purg (approved):

PR #155, opened 2022-10-31
Add background state as first state of recipe.

mattbrown11:
Update recipe so that the first step is always background.

### DIFF
--- a/config/tasks/task_steps_config-recipe_coffee.yaml
+++ b/config/tasks/task_steps_config-recipe_coffee.yaml
@@ -22,139 +22,146 @@ steps:
   # - std_conf: standard deviation of confidence for both true and false
   #             examples.
   - id: 0
+    activity_id: 0      # The first step must always be background.
+    description: >-
+      Background        # The first step must always be background.
+    median_duration_seconds: 5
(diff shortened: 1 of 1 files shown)


mattbrown11:
@Purg @joshanderson-kw I think I am currently the only consumer of this file, so this should be a low-impact change.

config/tasks/task_steps_config-recipe_coffee.yaml
```diff
@@ -22,182 +22,189 @@ steps:
   # - std_conf: standard deviation of confidence for both true and false
   #             examples.
   - id: 0
+    activity_id: 0      # The first step must always be background.
+    description: >-
+      Background        # The first step must always be background.
```
Purg, line 27: I'm not sure if the trailing comment on this line will behave the way you think it will. The `description` value that is loaded from this is `"Background        # The first step must always be background."`.

mattbrown11:
> 

No, during the init of ActivityHMMRos (https://github.com/PTG-Kitware/angel_system/blob/master/angel_system/activity_hmm/core.py#L20), it added that background state in automatically. So, it shouldn't appear different downstream.

PR #156, opened 2022-10-31
Dev/object detector speedup

alexander-lynch:


Purg:
This includes a commit that it should not: `Test reorganization of predict and device load order`

ros/angel_system_nodes/angel_system_nodes/object_detector_with_descriptors.py
```diff
@@ -14,6 +14,9 @@
 from angel_system.fasterrcnn.processing_utils import _get_image_blob
 from angel_utils.conversion import time_to_int
 
+import cProfile, pstats, io
+from pstats import SortKey
```
Purg, line 18: unused imports. remove.

PR #157, opened 2022-11-01
ActivityHMMRos can read mean and std coefficients from disk.

mattbrown11:


mattbrown11:
@joshanderson-kw @Purg

Purg (approved):

PR #158, opened 2022-11-01
Dev/add hand pose converter node 

joshanderson-kw:


scripts/hand_pose_converter.py
```diff
@@ -0,0 +1,153 @@
...
+# MediaFrameReader which provides the frames, instead of the Unity main camera.
+FOCAL_LENGTH_X = 1.6304
+FOCAL_LENGTH_Y = 2.5084
+
+PROJECTION_MATRIX = [
+    [FOCAL_LENGTH_X, 0.0, 0.0, 0.0],
+    [0.0, FOCAL_LENGTH_Y, 0.0, 0.0],
+    [0.0, 0.0, -1.0020020008087158, -0.20020020008087158],
+    [0.0, 0.0, -1.0, 0.0]
+]
```
Purg, line 27: I assume this could also use the content now "shared" in the angel_system package.

Purg (commented):
I think this looks good to me, at the moment. As discussed during stand-up, we're going to keep this on a side-branch from master until this 2D thread "pans out" fully.

PR #160, opened 2022-11-01
Fix bug in add_activity_classification adding too many states.

mattbrown11:


angel_system/activity_hmm/core.py
```diff
@@ -261,7 +261,8 @@ def add_activity_classification(self, label_vec, conf_vec, start_time,
             n += 1
             times_fill = np.linspace(last_time, start_time, n)
             times_fill = (times_fill[1:] + times_fill[:-1])/2
-            X_fill = np.tile(self.class_mean_conf, (len(times_fill), 1))
+
+            X_fill = np.zeros((len(times_fill), X.shape[1]))
```
Purg, line 265: Do you mean `self.X`?

PR #164, opened 2022-11-02
Added task complete vector output

Purg:
More specifically denotes which steps of the task are complete via index association to the task steps list in the message as well as retrieved from the `QueryTaskGraph` service.


bclipp-kitware (approved):

joshanderson-kw (approved):

joshanderson-kw:
We might need to rebuild the HL2 app after this since we're changing `TaskUpdate`

PR #165, opened 2022-11-03
Fix debug target enable and regenerate TaskUpdate msg

joshanderson-kw:


ovenmitt (approved):
Changes in FollowEyeTarget.cs look good.

PR #172, opened 2022-11-04
Fix bug with transitions jumping between background states.

mattbrown11:
@Purg 

angel_system/activity_hmm/core.py is the only consequential change to the live system, the rest are my testing scripts.


Purg:
this is basically the release branch now...

PR #173, opened 2022-11-04
Calculate precision and recall from confusion matrix

hdefazio:
- Changed precision and recall to be calculated from the confusion matrix
- Removed PR and ROC curves
- Added best/worst precision values output


Purg:
Is this PR still valid any more? Otherwise we could close it.

hdefazio:
Yeah, I think this is still valid. It looks like what I used to generate plots for the Nov slides

PR #175, opened 2023-02-06
Add HL2SS server plugin to ARUI unity project

joshanderson-kw:


Purg (approved):
Is there any versioning if the HL2SS plugin library?

joshanderson-kw:
> Is there any versioning if the HL2SS plugin library?

They do tagged releases in github. Is that what you mean?

Purg:
> > Is there any versioning if the HL2SS plugin library?
> 
> They do tagged releases in github. Is that what you mean?

I mean, where if anywhere do we locally record what version we are using? I'm not sure if that follows in their Unity plugin, but we are also using some manually imported materials that I didn't see any versioning info around. We should know what version we are using. It doesn't seem like there is anything we can do with segregating HL2SS things into their own version-encoded directories, so at the very least we need to add to our documentation what version we are making use of.

joshanderson-kw:
> > > Is there any versioning if the HL2SS plugin library?
> > 
> > 
> > They do tagged releases in github. Is that what you mean?
> 
> I mean, where if anywhere do we locally record what version we are using? I'm not sure if that follows in their Unity plugin, but we are also using some manually imported materials that I didn't see any versioning info around. We should know what version we are using. It doesn't seem like there is anything we can do with segregating HL2SS things into their own version-encoded directories, so at the very least we need to add to our documentation what version we are making use of.

I added the HL2SS version info to the top level README. Let me know if that should go somewhere else.

Also, I think we could remove the `HoloLens2-ResearchMode-Unity` folder now that we have the hl2ss plugin.

Purg (approved):

Purg:
> Also, I think we could remove the HoloLens2-ResearchMode-Unity folder now that we have the hl2ss plugin.

If that is not being utilized anymore, please make a branch/PR that removes it, making sure out app still builds appropriately.

PR #176, opened 2023-02-06
Add ROS HL2SS client node for RGB frame and hand joint data publishing

joshanderson-kw:
Depends on #175 

ROS node publishes PV camera frames as `Image` messages and hand joint pose data as `HandJointPoseUpdate` messages.


ros/angel_system_nodes/angel_system_nodes/hl2ss_ros_bridge.py
```diff
@@ -0,0 +1,316 @@
...
+        self._si_active = Event()
+        self._si_active.set()
+        self._si_thread = Thread(
+            target=self.si_publisher,
+            name="publish_frames"
+        )
+        self._si_thread.daemon = True
+        self._si_thread.start()
+        log.info("Starting publishing threads... Done")
+
```
Purg, line 110: extraneous newlines between methods. I/we need to add flake8 or black to CI this stuff... Generally there should be a single newline between intra-class methods.

ros/angel_system_nodes/angel_system_nodes/hl2ss_ros_bridge.py
```diff
@@ -0,0 +1,316 @@
...
+        # Operating mode
+        # 0: video
+        # 1: video + camera pose
+        # 2: query calibration (single transfer)
+        mode = hl2ss.StreamMode.MODE_1
+
+        # Camera parameters
+        width     = 1280
+        height    = 720
+        framerate = 30
```
Purg, line 125: This is fine for now, but we could expose these as node parameters with these values as defaults.

ros/angel_system_nodes/angel_system_nodes/hl2ss_ros_bridge.py
```diff
@@ -0,0 +1,316 @@
...
+
+        # Video encoding profile
+        profile = hl2ss.VideoProfile.H265_MAIN
+
+        # Encoded stream average bits per second
+        # Must be > 0
+        bitrate = 5*1024*1024
```
Purg, line 132: Is this a magic number? Otherwise what dictates this?

ros/angel_system_nodes/angel_system_nodes/hl2ss_ros_bridge.py
```diff
@@ -0,0 +1,316 @@
...
+        hand: str,
+        timestamp: int
+    ) -> HandJointPosesUpdate:
+        """
+        Extracts the hand joint poses data from the HL2SS SI structure
+        and forms a ROS HandJointPosesUpdate message.
+        """
+        # List containing joint names that matches the ordering in the HL2SS
+        # SI_HandJointKind class
+        joint_list = [
```
Purg, line 244: Part of me thinks this should be segregated out into a configuration file that this node is pointed to. Something lower bar would be to make this list a global variable instead of something that is created every time this method is called.

ros/angel_utils/python/angel_utils/conversion.py
```diff
@@ -33,6 +36,18 @@ def time_to_int(time_msg: Time) -> int:
     return (time_msg.sec * SEC_TO_NANO) + time_msg.nanosec
 
 
+def hl2ss_stamp_to_ros_time(timestamp: int) -> Time:
+    """
+    Convert the HL2SS timestamp which is an integer in hundreds of ns to a ROS
```
Purg, line 41: Hundreds of NS as in 1e7?

ros/angel_utils/python/angel_utils/conversion.py
```diff
@@ -33,6 +36,18 @@ def time_to_int(time_msg: Time) -> int:
...
+    Time message.
+    """
+    timestamp_sec = timestamp * HUN_NANO_TO_SEC
+    sec = math.floor(timestamp_sec)
+    nanosec = int((timestamp_sec - sec) * SEC_TO_NANO)
```
Purg, line 46: as in `HUN_NANO_TO_SEC` is 1e-7? If I understand correctly, we could simplify some of this math to
```
# Assuming HECTO_NS is defined as a global to int(1e7)
sec = timestamp // HECTO_NS
nanosec = timestamp % HECTO_NS
```

tmux/hl2ss_demo.yml
```diff
@@ -0,0 +1,79 @@
+#
+# System configuration to run the ANGEL system for the 2022/11 PI meeting and
+# Evaluation 1.
+#
+# This configuration is for the Coffee task/recipe.
```
Purg, line 5: This description comment could be updated.

tmux/hl2ss_demo.yml
```diff
@@ -0,0 +1,79 @@
...
+
+windows:
+  - hl2ss: ros2 run angel_system_nodes hl2ss_ros_bridge --ros-args
+      -r __ns:=${ROS_NAMESPACE}
+      -p ip_addr:=${HL2_IP}
+      -p image_topic:=PVFramesRGB
+      -p hand_pose_topic:=HandJointPoseData
+
+  # Visualize RGB Images being output from the headset
+  - rqt_2d_detections: rqt -s rqt_image_view/ImageView
```
Purg, line 57: This node name doesn't reflect what is being run under it.

tmux/hl2ss_demo.yml
```diff
@@ -0,0 +1,79 @@
...
+      --args ${ROS_NAMESPACE}/PVFramesRGB
+      --ros-args -p _image_transport:=raw
+  # Object detector
+  - object_detector: ros2 run angel_system_nodes object_detector_with_descriptors --ros-args
```
Purg, line 61: Running the object detector is overkill for demonstrating data stream output from the HL2SS interaction. If we had something to show/render the hand pose output, that would be better.

ros/angel_system_nodes/angel_system_nodes/hl2ss_ros_bridge.py
Purg: So, I would say at least refer to where you coped the "magic" from. This may be appropriate to be a global "static" variable. 5Mbps is a relatively small slice of a network connection pie (give our assumption of an isolated network), so I'm OK with leaving this hear if 5Mbps was already at the high end, especially since we are expecting to stream more than just the RGB frames (i.e. hand poses, grayscale cams eventually, etc.).

Purg:
Submoduling the hl2ss repo under the python package is interesting, but feels very odd to me. I don't have anything better to suggest, other than maybe having the repo submoduled outside of the angel_system package, and the angel system package having symlinks within in to reference using a relative path into the viewer directory of the submodule.

PR #177, opened 2023-02-08
BBN client ROS node

Purg:
This adds ROS2 messages and a node package that can translate top-level update messages into YAML, transmitting them over a parameterized TCP URL that is the receiving server.

This also has some minor-ish updates to some things like container build scripting modularization, internal pip version update to the latest, and updating the uncrustify config to be compatible with the latest version of that tool (0.76).

### DIFF
--- a/ros/bbn_integration/src/nodes/zmq_integration_client.cxx
+++ b/ros/bbn_integration/src/nodes/zmq_integration_client.cxx
@@ -1,5 +1,8 @@
+#include <chrono>
 #include <regex>
+#include <sstream>
 
+#include <bbn_integration/ros_to_yaml.h>
 #include <bbn_integration_msgs/msg/bbn_update.hpp>
 #include <rclcpp/rclcpp.hpp>
 #include <rclcpp_components/register_node_macro.hpp>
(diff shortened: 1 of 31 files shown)


ros/bbn_integration_msgs/msg/BBNCasualties.msg
```diff
@@ -0,0 +1,11 @@
+# 2a. CASUALTY COUNT / WORK STATIONS
```
joshanderson-kw, line 1: Were all of these BBN messages defined by BBN?

ros/bbn_integration/src/nodes/zmq_integration_client.cxx
```diff
@@ -0,0 +1,156 @@
...
+    // Override the transmission timestamp to "now"
+    auto const now = std::chrono::system_clock::now();
+    auto epoch_seconds =
+      std::chrono::duration< double >( now.time_since_epoch() ).count();
+    yaml_root[ "header" ][ "transmit timestamp" ] = epoch_seconds;
+
+    auto yaml_translation = YAML::Dump( yaml_root );
+
+    // Should return the number of bytes sent. If return does not have a value,
+    // then the zmq_send function returned code EAGAIN (11) "Try again".
```
joshanderson-kw, line 126: Should we try again here or just log the warning and continue?

ros/bbn_integration_msgs/msg/BBNCurrentSkill.msg
```diff
@@ -0,0 +1,5 @@
+# WHICH SKILL?
+string number
```
joshanderson-kw, line 2: Is this a number? Is there some mapping from skill to number somewhere?

ros/bbn_integration/src/nodes/zmq_integration_client.cxx
Purg: I'm not really sure. I think this should be a discussion with Brian V. and the BBN folks on what the "failure" plan is. I may need to read some more ZMQ docs. I am currently intentionally dropping the message on the floor and continuing instead of being "stuck" trying to send the same message (causing backup on the ROS side, which *will* drop messages).

Purg: I'm also not sure that EAGAIN "errors" are likely in our current client/server context: Neither the server nor the client (our thing) are using the `zmq::send_flags::dontwait` nor `zmq::recv_flags::dontwait` flags, which I think have to do with EAGAIN propagation.

joshanderson-kw: Yeah, it makes sense to not hold things up in the ROS side and get stuck in the callback function. That sounds good to me then.

PR #181, opened 2023-02-16
Delete old research mode folder

joshanderson-kw:
No longer used with HL2SS plugin.


Purg:
Just want a confirmation that our app still builds and operates nominally.

joshanderson-kw:
> Just want a confirmation that our app still builds and operates nominally.

Oh, right. I was able to build and run a quick test and see that the app runs normally.

Purg (approved):

PR #188, opened 2023-03-01
Handle user intent message in FeedbackGenerator

Purg:


ros/angel_system_nodes/angel_system_nodes/feedback_generator.py
```diff
@@ -26,18 +28,35 @@ class FeedbackGenerator(Node):
     def __init__(self):
         super().__init__(self.__class__.__name__)
 
-        self._activity_detector_topic = self.declare_parameter("activity_detector_topic", "ActivityDetections").get_parameter_value().string_value
-        self._object_detection_topic = self.declare_parameter("object_detection_topic", "ObjectDetections3d").get_parameter_value().string_value
-        # TODO: task needs to be updated to emit TaskNode
-        self._task_monitor_topic = self.declare_parameter("task_monitor_topic", "TaskUpdates").get_parameter_value().string_value
-        self._arui_update_topic = self.declare_parameter("arui_update_topic", "AruiUpdates").get_parameter_value().string_value
+        self._activity_detector_topic = (
```
joshanderson-kw, line 31: I think at some point you mentioned wanting to get away from using valid default topic names as the default parameter value, like `ActivityDetections`, so that we have to explicitly provide each argument to the node. Would that make sense here?

joshanderson-kw (approved):
Looks good! Just one minor comment/question.

ros/angel_system_nodes/angel_system_nodes/feedback_generator.py
Purg: Sure. I'm in here anyway.

PR #192, opened 2023-03-06
Handle user intent w/ ARUI bridge

joshanderson-kw:
Depends on #188


joshanderson-kw:
@ovenmitt, will you take a look at this? I modified the ARUI scripts a bit to pass the `InterpretedAudioUserIntentMsg` to the callback function, so we can publish it.

ros/angel_msgs/msg/SystemCommands.msg
```diff
@@ -9,6 +9,6 @@
 #
 
 # Task Monitor commands
-bool reset_current_task false
-bool previous_step      false
-bool next_step          false
+bool reset_current_task
+bool previous_step
+bool next_step
```
Purg, line 14: Did the default values not work, or cause something else to not work?

unity/ARUI/Assets/Scripts/ARUI/UIComponents/ConfirmationDialogue.cs
```diff
@@ -5,14 +5,22 @@
 using TMPro;
 using UnityEngine;
 using UnityEngine.Events;
+using RosMessageTypes.Angel;
+
+
+[System.Serializable]
+public class InterpretedAudioUserIntentEvent : UnityEvent<InterpretedAudioUserIntentMsg>
```
Purg, line 12: Could use a documenting comment for the purpose of the class.

unity/ARUI/Assets/Scripts/ARUI/TestData.cs
```diff
@@ -98,9 +102,10 @@ public void Update()
 
         if (Input.GetKeyUp(KeyCode.P))
         {
-            AngelARUI.Instance.SetUserIntentCallback(() => { AngelARUI.Instance.LogDebugMessage("The user confirmed the dialogue", true);  });
-            AngelARUI.Instance.TryGetUserFeedbackOnUserIntent("Do you need more information about the current task?");
-        }
+            AngelARUI.Instance.SetUserIntentCallback((intent) => { AngelARUI.Instance.LogDebugMessage("The user confirmed the dialogue", true);  });
+        InterpretedAudioUserIntentMsg intentMsg = new InterpretedAudioUserIntentMsg();
+        intentMsg.user_intent = "Do you need more information about the current task?";
+        AngelARUI.Instance.TryGetUserFeedbackOnUserIntent(intentMsg);        }
```
Purg, line 108: The indentation of this looks to need an additional level, and the trailing `}` looks to not be on its own line.

unity/ARUI/Assets/Scripts/ARUI/AngelARUI.cs
```diff
@@ -97,24 +99,24 @@ public void SetTasks(string[,] tasks)
...
 
     /// <summary>
-    /// If confirmation action is set - SetUserIntentCallback(...) - and no confirmation window is active at the moment, the user is shown a 
-    /// timed confirmation window. Recommended text: "Did you mean ...". If the user confirms the dialogue, the onUserIntentConfirmedAction action is invoked. 
+    /// If confirmation action is set - SetUserIntentCallback(...) - and no confirmation window is active at the moment, the user is shown a
+    /// timed confirmation window. Recommended text: "Did you mean ...". If the user confirms the dialogue, the onUserIntentConfirmedAction action is invoked.
     /// </summary>
     /// <param name="msg">message that is shown in the confirmation dialogue</param>
-    public void TryGetUserFeedbackOnUserIntent(string msg)
+    public void TryGetUserFeedbackOnUserIntent(InterpretedAudioUserIntentMsg intentMsg)
```
Purg, line 109: Ideally, we would modularize the ARUI component from our ROS-specific semantics (e.g. the trans-piled message classes) so that the ARUI could be used by other TA1 members, however if this first demonstration requires this, then we may be able to make due.

A next step may be for there to be a C# class whose properties reflect what our ROS-message has so that we have a 1-to-1 correspondence between the class and the message, but the class is part of the ARUI API and is more transportable.

unity/ARUI/Assets/Scripts/ARUI/AngelARUI.cs
```diff
@@ -97,24 +99,24 @@ public void SetTasks(string[,] tasks)
...
     {
         if (onUserIntentConfirmedAction == null || confirmationWindow != null) return;
 
         GameObject window = Instantiate(confirmationWindowPrefab, transform);
         confirmationWindow = window.AddComponent<ConfirmationDialogue>();
-        confirmationWindow.InitializeConfirmationNotification(msg, onUserIntentConfirmedAction);
+        confirmationWindow.InitializeConfirmationNotification(intentMsg, onUserIntentConfirmedAction);
```
Purg, line 115: Maybe a @ovenmitt question: What happens if multiple user intents come in in quick succession such that this method (`TryGetUserFeedbackOnUserIntent`) is invoked again before the user has finished with the first confirmation? Are multiple confirmation windows created for the user to interact with?

unity/ARUI/Assets/Scripts/hl2ss.cs
```diff
@@ -7,6 +7,22 @@
 
 public class hl2ss : MonoBehaviour
 {
+    private Logger _logger = null;
+
+    /// <summary>
+    /// Lazy acquire the ARUI logger object and return the reference to it.
+    /// </summary>
+    /// <returns>Logger instance reference.</returns>
+    private ref Logger logger()
```
Purg, line 16: Does this want/need any thread-safe protections? I'm assuming you tested this and didn't have any issues, so I may be prematurely concerned.

unity/ARUI/Assets/Scripts/ARUI/UIComponents/ConfirmationDialogue.cs
ovenmitt: Yes, I am currently working on this class's documentation in my fork.

unity/ARUI/Assets/Scripts/ARUI/AngelARUI.cs
ovenmitt: Right now, it is not designed for multiple user intents at the same time because, from the user's perspective, that is not something useful. Imagine you are talking to Siri, or any other person, really. There is usually one user intent at a time. 

Since it is _technically_ possible to get 'misfires' or 'multiple detected user intent within a short time', something in the system would have to make decisions about what user intent is the most important (or the correct one) and communicate that one to the user.

Naively, I could just store all incoming intents in a queue and show them the user in this order.

unity/ARUI/Assets/Scripts/ARUI/AngelARUI.cs
ovenmitt: I agree, I would like to keep the ARUI independent (if possible), so it can be used by other systems, but we can leave the changes in the AngelARUI script for the demo.

ovenmitt (commented):
Changing the AngelARUI script is not ideal, but okay for the demo.

unity/ARUI/Assets/Scripts/ARUI/AngelARUI.cs
Purg: OK, thanks for the input. Let's see how this implementation performs in combination with the audio intent components before we (I) get prematurely concerned about capability.

unity/ARUI/Assets/Scripts/ARUI/UIComponents/ConfirmationDialogue.cs
Purg: I appreciate it! This instance is more for @joshanderson-kw with this new class that was introduced. I think it's likely just a simple sentence to add.

ros/angel_msgs/msg/SystemCommands.msg
joshanderson-kw: Yeah, they were causing compiler error when I tried to generate the C# scripts. It seemed like that the ROS Unity library didn't support the default syntax.

unity/ARUI/Assets/Scripts/hl2ss.cs
joshanderson-kw: Yeah, I tested this and it seemed OK. We only log the IP address at the beginning, so hopefully there aren't any threading issues there.

PR #193, opened 2023-03-06
Add user intent to cmd ROS node

joshanderson-kw:
Converts explicit and confirmed user intent messages to system command messages.


ros/angel_system_nodes/angel_system_nodes/intent_to_command.py
```diff
@@ -0,0 +1,134 @@
...
+                "Please provide the confirmed intent topic with the `confirmed_intent_topic` parameter"
+            )
+        if self._intent_to_cmd_map_path == "":
+            raise ValueError(
+                "Please provide the intent command map path with the `intent_to_cmd_map` parameter"
+            )
+        if self._sys_cmd_topic == "":
+            raise ValueError(
+                "Please provide the system command topic with the `sys_cmd_topic` parameter"
+            )
```
Purg, line 55: I attempted to be more concise about this in my other branch. Would it make sense to make use of that here, or do you think that is extraneous?

ros/angel_system_nodes/angel_system_nodes/intent_to_command.py
```diff
@@ -0,0 +1,134 @@
...
+            self._explicit_intent_topic,
+            self.intent_callback,
+            1
+        )
+        self._subscription = self.create_subscription(
+            InterpretedAudioUserIntent,
+            self._confirmed_intent_topic,
+            self.intent_callback,
+            1
+        )
```
Purg, line 79: Maybe these topics don't need to be separated. We could probably just have one topic that inputs to this that both the "explicit" and "confirmed" sources publish to.

ros/angel_system_nodes/angel_system_nodes/intent_to_command.py
```diff
@@ -0,0 +1,134 @@
...
+                f"User intent `{intent.user_intent}` not found in sys cmd map. "
+                f"Available intents are {list(self._intent_to_cmd_map.keys())}"
+            )
+            return
+
+        sys_cmd_msg = SystemCommands()
+
+        try:
+            # Set the corresponding sys cmd field to True
+            setattr(sys_cmd_msg, sys_cmd, True)
```
Purg, line 107: This is OK for now since we just have a few boolean-based commands, but this will break down when we have anything more complex. I'm not sure I have a concrete recommendation other than adding a comment before this that acknowledges this.

ros/angel_system_nodes/angel_system_nodes/intent_to_command.py
Purg: Did that alternative approach seem any easier/better to you? I want to get more than my thoughts into the mix.

PR #197, opened 2023-03-09
Add mock ASR node and User Intent Detection Node

derkmed:
This is a prototype for ASR and User Intent Detection based on ASR text.

The proposed ASR node is dependent on a Columbia machine running at "http://communication.cs.columbia.edu:8058/home". Please ping @derkmed if there are any issues connecting to it.

An upcoming PR will refactor the IntentDetector node to leverage the [InterpretedAudioUserIntent message](https://github.com/PTG-Kitware/angel_system/blob/master/ros/angel_msgs/msg/InterpretedAudioUserIntent.msg). Meanwhile, we will remove the IntentDetection.msg added in this PR.

DO_NOT_SUBMIT

### DIFF
--- a/ros/angel_msgs/msg/Utterance.msg
+++ b/ros/angel_msgs/msg/Utterance.msg
@@ -2,8 +2,5 @@
 # Represents an utterance from ASR.
 #
 
-# Speaker of the utterance.
-string speaker
-
 # Utterance value.
 string value
--- a/ros/angel_system_nodes/angel_system_nodes/asr.py
+++ b/ros/angel_system_nodes/angel_system_nodes/asr.py
@@ -5,11 +5,12 @@ import threading
 import time
 import wave
 
-from angel_msgs.msg import HeadsetAudioData, Utterance
 from nltk.tokenize import sent_tokenize
 import rclpy
 from rclpy.node import Node
 import simpleaudio as sa
+
+from angel_msgs.msg import HeadsetAudioData, Utterance
 
 
 AUDIO_TOPIC = "audio_topic"
@@ -150,7 +151,6 @@ class ASR(Node):
             self.log.info("Complete ASR text is:\n" + f"\"{response.text}\"")
             for sentence in sent_tokenize(response.text):
                 utterance_msg = Utterance()
-                utterance_msg.speaker = "user"
                 utterance_msg.value = sentence
                 self.log.info("Publishing message: " + f"\"{sentence}\"")
                 self._publisher.publish(utterance_msg)
--- a/ros/angel_system_nodes/angel_system_nodes/asr.py
+++ b/ros/angel_system_nodes/angel_system_nodes/asr.py
@@ -1,20 +1,15 @@
 import pyaudio
-import rclpy
-from rclpy.node import Node
 import requests
-import simpleaudio as sa
 import tempfile
 import threading
 import time
 import wave
 
 from angel_msgs.msg import HeadsetAudioData, Utterance
-
-import nltk
 from nltk.tokenize import sent_tokenize
-
-
-nltk.download('punkt')
+import rclpy
+from rclpy.node import Node
+import simpleaudio as sa
 
 
 AUDIO_TOPIC = "audio_topic"
--- a/ros/angel_system_nodes/angel_system_nodes/asr.py
+++ b/ros/angel_system_nodes/angel_system_nodes/asr.py
@@ -136,24 +136,31 @@ class ASR(Node):
     def asr_server_request_thread(self, audio_data, num_channels, sample_rate):
         with self.asr_server_lock:
 
-            temp_file = tempfile.NamedTemporaryFile(suffix='.wav', delete=False)
+            temp_file = tempfile.NamedTemporaryFile(suffix='.wav', delete=True)
             self.log.info(f"Writing audio to temporary file: {temp_file.name}")
-            wf = wave.open(temp_file, 'wb')
-            wf.setnchannels(num_channels)
+            wav_write = wave.open(temp_file, 'wb')
+            wav_write.setnchannels(num_channels)
             # TODO (derekahmed) We should figure out how this value was derived
             # and make this a constant accordingly.
-            wf.setsampwidth(4) # Derived from audio_player.py
-            wf.setframerate(sample_rate)
-            wf.writeframes(audio_data)
-            with open(temp_file.name, 'rb') as f:
-                response = requests.post(self._asr_server_url, files={'audio_data': f})
+            wav_write.setsampwidth(4) # Derived from audio_player.py
+            wav_write.setframerate(sample_rate)
+            wav_write.writeframes(audio_data)
 
-            self.log.info("Complete ASR text is:\n" + f"\"{response.text}\"")
-            for sentence in sent_tokenize(response.text):
-                utterance_msg = Utterance()
-                utterance_msg.value = sentence
-                self.log.info("Publishing message: " + f"\"{sentence}\"")
-                self._publisher.publish(utterance_msg)
+            with open(temp_file.name, 'rb') as temp_file:
+                response = requests.post(self._asr_server_url,
+                                         files={'audio_data': temp_file})
+            temp_file.close()
+            
+            if not response.ok:
+                self.log.error("ASR Server Response contains an error.")
+                return
+            if response:
+                self.log.info("Complete ASR text is:\n" + f"\"{response.text}\"")
+                for sentence in sent_tokenize(response.text):
+                    utterance_msg = Utterance()
+                    utterance_msg.value = sentence
+                    self.log.info("Publishing message: " + f"\"{sentence}\"")
+                    self._publisher.publish(utterance_msg)
 
 
 def main():
--- a/ros/angel_system_nodes/angel_system_nodes/asr.py
+++ b/ros/angel_system_nodes/angel_system_nodes/asr.py
@@ -3,6 +3,7 @@ import rclpy
 from rclpy.node import Node
 import requests
 import simpleaudio as sa
+import tempfile
 import threading
 import time
 import wave
@@ -11,23 +12,62 @@ from angel_msgs.msg import HeadsetAudioData, Utterance
 
 import nltk
 from nltk.tokenize import sent_tokenize
+
+
 nltk.download('punkt')
 
 
+AUDIO_TOPIC = "audio_topic"
+UTTERANCES_TOPIC = "utterances_topic"
+ASR_SERVER_URL = "asr_server_url"
+ASR_REQ_SEGMENT_SECONDS_DURATION = "asr_req_segment_duration"
+
+
 class ASR(Node):
+
     def __init__(self):
         super().__init__(self.__class__.__name__)
+        self.log = self.get_logger()
 
-        self.declare_parameter("audio_topic", "HeadsetAudioData")
-        self.declare_parameter("utterances_topic", "utterances_topic")
-        self.declare_parameter("asr_server_url", "http://communication.cs.columbia.edu:8058/home")
-        self.declare_parameter("tmp_wav_filename", "/tmp/asr.wav")
+        parameter_names = [
+            AUDIO_TOPIC,
+            UTTERANCES_TOPIC,
+            ASR_SERVER_URL,
+            ASR_REQ_SEGMENT_SECONDS_DURATION
+        ]
+        set_parameters = self.declare_parameters(
+            namespace="",
+            parameters=[(p,) for p in parameter_names],
+        )
+        # Check for not-set parameters.
+        some_not_set = False
+        for p in set_parameters:
+            if p.type_ is rclpy.parameter.Parameter.Type.NOT_SET:
+                some_not_set = True
+                self.log.error(f"Parameter not set: {p.name}")
+        if some_not_set:
+            raise ValueError("Some parameters are not set.")
         
-
-        self._audio_topic = self.get_parameter("audio_topic").get_parameter_value().string_value
-        self._utterances_topic = self.get_parameter("utterances_topic").get_parameter_value().string_value
-        self._asr_server_url = self.get_parameter("asr_server_url").get_parameter_value().string_value
-        self._tmp_wav_filename = self.get_parameter("tmp_wav_filename").get_parameter_value().string_value
+        self._audio_topic =\
+            self.get_parameter(AUDIO_TOPIC).get_parameter_value().string_value
+        self._utterances_topic =\
+            self.get_parameter(UTTERANCES_TOPIC).get_parameter_value().string_value
+        self._asr_server_url =\
+            self.get_parameter(ASR_SERVER_URL).get_parameter_value().string_value
+        self.segmentation_duration =\
+            self.get_parameter(ASR_REQ_SEGMENT_SECONDS_DURATION).value
+        self.log.info(f"Audio topic: "
+                      f"({type(self._audio_topic).__name__}) "
+                      f"{self._audio_topic}")
+        self.log.info(f"Utterances topic: "
+                      f"({type(self._utterances_topic).__name__}) "
+                      f"{self._utterances_topic}")
+        self.log.info(f"Columbia ASR Server URL: "
+                      f"({type(self._asr_server_url).__name__}) "
+                      f"{self._asr_server_url}")
+        self.log.info(f"Columbia ASR Server Request Segmentation Duration: "
+                      f"({type(self.segmentation_duration).__name__}) "
+                      f"{self.segmentation_duration}")
 
         self.subscription = self.create_subscription(
             HeadsetAudioData,
@@ -35,7 +75,7 @@ class ASR(Node):
             self.listener_callback,
             1)
         
-        # TODO(derekahmed): Add internal queueing to reduce publisher queue
+        # TODO(derekahmed): Add internal buffering to reduce publisher queue
         # size to 1.
         self._publisher = self.create_publisher(
             Utterance,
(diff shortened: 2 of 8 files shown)


joshanderson-kw:
Could the Intent Detector and the ASR nodes be combined into one node? So that the ASR node just publishes `InterpretedAudioUserIntent` messages directly, instead of going from `Utterance` to `InterpretedAudioUserIntent` via the Intent detector node. Or, maybe I misunderstanding how the two nodes interact.

derkmed:
We _could_ consolidate them into a single node. But, I believe it makes sense decoupling them as different nodes, especially if we have other dialogue-related classifiers that might consume the ASR text utterances. Btw, the Utterance message is likely subject to modifications down-the-line.

Though, I'm not entirely familiar with any resource limitations, which could be an entirely deciding reason for doing what you proposed. If your concern is adding another message, we could always reuse the `InterpretedAudioUserIntent` instead of introducing `Utterance`, but such an intentional misuse of an entity could lead to future confusion or bugs down the line I think. Happy to discuss over a call.

derkmed:
BTW please see the new refactoring of the IntentDetection Node. I can apply any review comments/suggestions/requests.

ros/angel_system_nodes/angel_system_nodes/intent_detector.py
```diff
@@ -0,0 +1,140 @@
...
+import rclpy
+from rclpy.node import Node
+from std_msgs.msg import Header
+
+from angel_msgs.msg import InterpretedAudioUserIntent, Utterance
+
+# Please refer to labels defined in
+# https://docs.google.com/document/d/1uuvSL5de3LVM9c0tKpRKYazDxckffRHf7IAcabSw9UA .
+NEXT_STEP_KEYPHRASES = ['skip', 'next step']
+PREV_STEP_KEYPHRASES = ['prev step', 'last step', 'go back']
```
joshanderson-kw, line 16: Maybe could have these keyphrases as a config file passed into this node. 

Locally, I added `next`, `previous`, and `previous step`.

ros/angel_system_nodes/angel_system_nodes/asr.py
```diff
@@ -0,0 +1,136 @@
...
+            10
+        )
+
+        self.audio_stream = bytearray()
+        self.audio_stream_lock = threading.RLock()
+        self.asr_server_lock = threading.RLock()
+        self.t = threading.Thread()
+
+        self.audio_stream_duration = 0.0
+        self.segmentation_duration = 30.0
```
joshanderson-kw, line 52: Could make this is a ROS parameter we pass into this node.

ros/angel_system_nodes/package.xml
```diff
@@ -17,6 +17,7 @@
   <depend>image_transport</depend>
   <depend>cv_bridge</depend>
   <depend>python3-opencv</depend>
+  <depend>python3-pyaudio</depend>
```
joshanderson-kw, line 20: I think we'll need to add the pyaudio dependency to the pyproject list. I also got an import error after pip install pyaudio, so I had to do `apt install python3-pyaudio`.

I think we can add that apt install to the Dockerfile here: https://github.com/PTG-Kitware/angel_system/blob/master/docker/workspace-build/Dockerfile#L20

@Purg, is that the right place to install pyaudio?

ros/angel_system_nodes/angel_system_nodes/intent_detector.py
```diff
@@ -0,0 +1,140 @@
...
+        lower_utterance = msg.value.lower()            
+        if self.contains_phrase(lower_utterance, NEXT_STEP_KEYPHRASES):
+            intent_msg.user_intent = LABELS[0]
+            intent_msg.confidence = 0.5
+        elif self.contains_phrase(lower_utterance, PREV_STEP_KEYPHRASES):
+            intent_msg.user_intent = LABELS[1]
+            intent_msg.confidence = 0.5
+        else:
+            intent_msg.user_intent = LABELS[2]
+            intent_msg.confidence = 0.5
```
joshanderson-kw, line 110: If the utterance does not contain any of the keyphrases, I think we could ignore it and return immediately.

I added the return statement locally, and it helped the ARUI not get bogged down with "Other" messages to confirm.

joshanderson-kw (commented):
Tested locally within the live ANGEL system and it works pretty well! Just a few small comments.

ros/angel_system_nodes/package.xml
Purg: I think I recommended @derkmed to add this here. If this dep is *only* used in the ROS node space, then this is the correct place to add this dep. The pyproject file encapsulates the deps only for the `angel_system` python package at the root level (NOT the `./ros/` stuff).

ros/angel_system_nodes/angel_system_nodes/asr.py
```diff
@@ -0,0 +1,136 @@
...
+        # size to 1.
+        self._publisher = self.create_publisher(
+            Utterance,
+            self._utterances_topic,
+            10
+        )
+
+        self.audio_stream = bytearray()
+        self.audio_stream_lock = threading.RLock()
+        self.asr_server_lock = threading.RLock()
```
Purg, line 48: It would be good to explicitly comment which properties these locks control. E.g. `audio_stream_lock` protects `audio_stream`, `audio_stream_duration`, `t`, *and* `prev_timestamp`.

ros/angel_system_nodes/angel_system_nodes/asr.py
```diff
@@ -0,0 +1,136 @@
...
+
+    asr = ASR()
+    
+    rclpy.spin(asr)
+    
+    asr.destroy_node()
+    rclpy.shutdown()
+
+if __name__ == '__main__':
+    main()
```
Purg, line 136: Nitpick: Newline at end of file.

ros/angel_system_nodes/angel_system_nodes/intent_detector.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+UTTERANCES_TOPIC = "utterances_topic"
+PARAM_EXPECT_USER_INTENT_TOPIC = "expect_user_intent_topic"
+PARAM_INTERP_USER_INTENT_TOPIC = "interp_user_intent_topic"
+
+class IntentDetector(Node):
+    '''
+    As of Q12023, intent detection is derived heuristically. This will be shifted
+    to a model-based approach in the neat-future.
```
Purg, line 35:
```suggestion
    to a model-based approach in the near-future.
```

ros/angel_system_nodes/angel_system_nodes/intent_detector.py
```diff
@@ -0,0 +1,140 @@
...
+from angel_msgs.msg import InterpretedAudioUserIntent, Utterance
+
+# Please refer to labels defined in
+# https://docs.google.com/document/d/1uuvSL5de3LVM9c0tKpRKYazDxckffRHf7IAcabSw9UA .
+NEXT_STEP_KEYPHRASES = ['skip', 'next step']
+PREV_STEP_KEYPHRASES = ['prev step', 'last step', 'go back']
+OVERRIDE_KEYPHRASES = ['angel', 'angel system']
+
+# TODO(derekahmed): Please figure out how to keep this sync-ed with
+# config/angel_system_cmds/user_intent_to_sys_cmd_v1.yaml.
```
Purg, line 20: Inline-thought: A possible method is checking set intersection with the keys in that command conversion config (maybe minus the "Other" label here for your failure catch).

ros/angel_system_nodes/setup.py
```diff
@@ -40,3 +42,4 @@
             ],
     },
 )
+1
```
Purg, line 45: Is this an accidental "1" value at the end of the file?

tmux/eval_asr.yml
```diff
@@ -0,0 +1,67 @@
+#
+# Used to create ROS bags for all topics published from the HL2.
+# This configuration should be run by itself (e.g. not in combination with
+# another tmuxinator launch).
```
Purg, line 4: This comment could use updating.

ros/angel_system_nodes/package.xml
Purg: I also forgot to add that we do not need to add an apt install to the dockerfile as this dep is slurped in via the rosdep component of the build that is part of the container. This step effectively performers additional ros-package specific dep declaration (like here).

tmux/eval_asr.yml
derkmed: Will resolve.

tmux/eval_asr.yml
```diff
@@ -0,0 +1,67 @@
...
+# Controls whether the tmux session should be attached to automatically. Defaults to true.
+# attach: false
+
+windows:
+  # - ros_bag_play: ros2 bag play <<PATH_TO_BAG_FILE>>
+  - ros_bag_play: ros2 bag play /angel_workspace/ros_bags/rosbag2_2023_03_01-17_28_00/rosbag2_2023_03_01-17_28_00_0.db3
+  - asr: ros2 run angel_system_nodes asr --ros-args
+      -r __ns:=${ROS_NAMESPACE}
+      -p audio_topic:=HeadsetAudioData
+      -p utterance_topic:=utterances_topic
```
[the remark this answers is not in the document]
derkmed: Will resolve.

ros/angel_system_nodes/setup.py
derkmed: Will resolve.

ros/angel_system_nodes/angel_system_nodes/intent_detector.py
derkmed: Will resolve.

ros/angel_system_nodes/angel_system_nodes/asr.py
derkmed: Will resolve.

ros/angel_system_nodes/angel_system_nodes/asr.py
```diff
@@ -0,0 +1,136 @@
...
+
+class ASR(Node):
+    def __init__(self):
+        super().__init__(self.__class__.__name__)
+
+        self.declare_parameter("audio_topic", "HeadsetAudioData")
+        self.declare_parameter("utterances_topic", "utterances_topic")
+        self.declare_parameter("asr_server_url", "http://communication.cs.columbia.edu:8058/home")
+        self.declare_parameter("tmp_wav_filename", "/tmp/asr.wav")
+
```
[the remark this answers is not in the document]
derkmed: Will resolve. Also had not realized PEP8 was the style guide and will try to apply accordingly.

ros/angel_system_nodes/angel_system_nodes/asr.py
```diff
@@ -0,0 +1,136 @@
...
+nltk.download('punkt')
+
+
+class ASR(Node):
+    def __init__(self):
+        super().__init__(self.__class__.__name__)
+
+        self.declare_parameter("audio_topic", "HeadsetAudioData")
+        self.declare_parameter("utterances_topic", "utterances_topic")
+        self.declare_parameter("asr_server_url", "http://communication.cs.columbia.edu:8058/home")
```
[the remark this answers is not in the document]
derkmed: I agree that it's something that should be internal. @maxlchen had set this server up to make it easier on our (Columbia) end for development & research. 

We have a non-production ready Flask server up and running which we can add you to later.

ros/angel_system_nodes/angel_system_nodes/asr.py
derkmed: Will resolve.

ros/angel_system_nodes/angel_system_nodes/asr.py
derkmed: Will resolve.

ros/angel_msgs/msg/Utterance.msg
```diff
@@ -0,0 +1,9 @@
+#
+# Represents an utterance from ASR.
+#
+
+# Speaker of the utterance.
+string speaker
+
+# Utterance value.
+string value
```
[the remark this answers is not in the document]
derkmed: Yes. In hindsight, this may be a byproduct of over-engineering early-on. I can remove the speaker field.

angel_system/interfaces/emotion_models.py
```diff
@@ -0,0 +1,75 @@
+import abc
```
derkmed, line 1: @maxlchen are these emotion files something we want in this PR? These were in the forked repo when I started and I suspect the answer is no.

angel_system/impls/emotion_models/ComParE_model.py
```diff
@@ -0,0 +1,53 @@
+import importlib.util
```
derkmed, line 1: @maxlchen are these emotion files something we want in this PR? These were in the forked repo when I started and I suspect the answer is no.

ros/angel_system_nodes/angel_system_nodes/intent_detector.py
derkmed: Done.

ros/angel_msgs/msg/Utterance.msg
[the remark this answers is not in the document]
derkmed: Removed.

ros/angel_system_nodes/angel_system_nodes/intent_detector.py
derkmed: Do you have an example on the convention to follow for node configs? 

The one concern for investing effort into that is that entire node implementation is only an interim hacky fix. The end goal is a classifier that would entail us removing these keyphrases altogether. If it'll make things go more smoothly in the demo, I can try setting up a config.

joshanderson-kw: You could take a look at the config for the intent_to_cmd node, which uses this simple yaml config https://github.com/PTG-Kitware/angel_system/blob/master/config/angel_system_cmds/user_intent_to_sys_cmd_v1.yaml. Not super high priority though.

I would definitely add "previous" and "next" to the key phrases list though, if you can't get to the config.

ros/angel_system_nodes/package.xml
joshanderson-kw: Ah, sorry, ignore me then. I didn't do a new docker container build so maybe that's why I was getting import errors.

ros/angel_system_nodes/angel_system_nodes/asr.py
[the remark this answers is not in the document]
Purg: Yeah, that was a communication gap on our end (not helped with lack of CI). I believe it is generally assumed with python that when there is a lack of any specific style guide, PEP8 applies.

unity/ARUI/Assets/MixedRealityToolkit.Generated/ProjectPreferences.asset
```diff
@@ -24,7 +24,7 @@ MonoBehaviour:
     keys:
     - _MixedRealityToolkit_Editor_ConfiguratorState
     - MixedRealityToolkit_Editor_AudioSpatializerCount
-    values: f401000002000000
+    values: c800000002000000
```
Purg, line 27: What preference change does this reflect? If unknown, I'm not sure we should include the change.

derkmed: Ah, that shouldn't have been added. Will resolve.

angel_system/impls/emotion_models/ComParE_model.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+    def __call__(
+      self,
+      X: Iterable[np.ndarray],
+      sampling_rate: int
+    ) -> int:
+        """
+        Calls `classify_emotion() with the given iterable modality representation.`
+        """
```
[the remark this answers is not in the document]
derkmed: File has been removed.

angel_system/impls/emotion_models/ComParE_model.py
```diff
@@ -0,0 +1,53 @@
+import importlib.util
+import logging
+from typing import Tuple, Iterable, Dict, Hashable, List, Union, Any
+from types import MethodType
+import json
+import numpy as np
+from joblib import load
+import opensmile
+
```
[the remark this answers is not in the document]
derkmed: File removed.

angel_system/impls/emotion_models/ComParE_model.py
```diff
@@ -0,0 +1,53 @@
...
+    def __call__(
+      self,
+      X: Iterable[np.ndarray],
+      sampling_rate: int
+    ) -> int:
+        """
+        Calls `classify_emotion() with the given iterable modality representation.`
+        """
+        return self.classify(X)
```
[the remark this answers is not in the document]
derkmed: File removed.

ros/angel_system_nodes/angel_system_nodes/asr.py
```diff
@@ -0,0 +1,136 @@
...
+import simpleaudio as sa
+import threading
+import time
+import wave
+
+from angel_msgs.msg import HeadsetAudioData, Utterance
+
+import nltk
+from nltk.tokenize import sent_tokenize
+nltk.download('punkt')
```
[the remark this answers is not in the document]
derkmed: Followed up offline.

Will add `RUN python -m nltk.downloader punkt` command after the "workspace_build_pydep_install" part of the docker/workspace-build/Dockerfile, as advised by https://stackoverflow.com/questions/31143015/docker-nltk-download .

ros/angel_system_nodes/angel_system_nodes/asr.py
```diff
@@ -88,37 +128,31 @@ def listener_callback(self, msg):
...
-    def asr_server_request_thread(self, audio_data, num_channels, 
-        sample_rate, wav_output_filename):
-        log = self.get_logger()
+    def asr_server_request_thread(self, audio_data, num_channels, sample_rate):
         with self.asr_server_lock:
 
-            # TODO(@derekahmed): We should use a temporary audio file instead. Otherwise,
-            # we should at least delete the saved audio file.
-            wf = wave.open(wav_output_filename, 'wb')
+            temp_file = tempfile.NamedTemporaryFile(suffix='.wav', delete=False)
```
Purg, line 139: @derkmed Was this set to `delete=False` for a reason? It seems like this should be `True` instead so that we do not accumulate temporary audio files.

derkmed: Resolved.

Purg (approved):

PR #200, opened 2023-03-13
View Management Intergration and Improvements NLI 2.0

ovenmitt:
Includes everything from the closed PR + a hotfix that would cause app to crash.

-  Improvement confirmation dialogue (top of FOV, instead of the bottom, added audio feedback and touch input)
- Added view management (for orb (controllable), tasklist, confirmation dialogue and hands (all non-controllables). The objective of view management is to avoid decreasing the legibility of virtual or real objects in the scene. Controllable will move away from non-controllable obejcts (e.g., the orb should not overlap with hands if the user is busy working on a task)
- Minor improvements (task list fixed with transparent items, fixed orb message not shown when looking at task list)
- Added 'stop' keyword that immediately stops the audio task instructions
- Audio task instructions can be muted
-  Code documentation and clean-up


joshanderson-kw:
Bettina and I just tested this and it looks good! There are a few minor changes I made that I will make a new PR for after this is merged.

PR #204, opened 2023-03-15
Fix checkmark issues with berkeley task monitor

hdefazio:
- Fixed checkmarks not properly functioning at the beginning/end of the list (now able to move to the beginning - nothing checked and the end - everything checked without issue)
- Only update checkmarks when we send a TaskUpdate (this allows the manual checkmarks to not be immediately overwritten)


Purg:
@hdefazio Could you add a summary of what this changed? I don't remember any more.

PR #207, opened 2023-04-04
Update to use latest HL2SS APi

joshanderson-kw:
Update HL2SS submodule to latest commit, which changes the image payload call slightly.


joshanderson-kw:
Closing in favor of #208

PR #214, opened 2023-04-12
Publish HL2SS spatial map

joshanderson-kw:


ros/angel_system_nodes/angel_system_nodes/hl2ss_ros_bridge.py
```diff
@@ -77,63 +92,69 @@ class HL2SSROSBridge(Node):
...
-        self.pv_height = (
-            self.declare_parameter("pv_height", 720)
-            .get_parameter_value()
-            .integer_value
-        )
-        self.pv_framerate = (
-            self.declare_parameter("pv_framerate", 30)
-            .get_parameter_value()
-            .integer_value
+        self.log = self.get_logger()
```
Purg, line 95: Possibly yes, but I wasn't sure if the logging instance returned by `get_logger()` was thread-safe.

joshanderson-kw: Oh, hadn't thought of that... I went ahead and reverted that.

ros/angel_system_nodes/angel_system_nodes/hl2ss_ros_bridge.py
```diff
@@ -365,15 +422,95 @@ def audio_publisher(self) -> None:
...
+                    m_v.y = float(v[1])
+                    m_v.z = float(v[2])
+                    mesh_shape.vertices.append(m_v)
+
+                spatial_mesh_msg.mesh = mesh_shape
+
+                # Publish!
+                self.ros_sm_publisher.publish(spatial_mesh_msg)
+
+            time.sleep(5)
```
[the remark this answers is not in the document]
joshanderson-kw: Done

ros/angel_system_nodes/angel_system_nodes/hl2ss_ros_bridge.py
```diff
@@ -161,6 +194,8 @@ def __init__(self):
         log.info("SI client connected!")
         self.connect_hl2ss_audio()
         log.info("Audio client connected!")
+        self.connect_hl2ss_sm()
+        log.info("SM client connected!")
```
Purg, line 198: Do you know what the behavior is if the HL2SS server does not have the Spatial-mapping functionality enabled? My concern is compatibility with the BBN system which I believe is not outputting all of the possible streams (or is fully up to date). If this does not gracefully fail, we will likely need to add some ability here to "disable" trying to connect to various servers (e.g. if "disable" is passed in as the "topic", don't create a client, etc.).

Purg (approved):
LGTM. Only possibly concern in the comment is compatibility with the BBN system which is the near-term push.

ros/angel_system_nodes/angel_system_nodes/hl2ss_ros_bridge.py
joshanderson-kw: This is added now, so if you pass "disable" as the topic name for any of the sensor streams, they will be disabled.

Purg (approved):

PR #215, opened 2023-04-13
Add headset pose publishing

joshanderson-kw:
Depends on #214

### DIFF
--- a/ros/angel_system_nodes/angel_system_nodes/hl2ss_ros_bridge.py
+++ b/ros/angel_system_nodes/angel_system_nodes/hl2ss_ros_bridge.py
@@ -5,7 +5,6 @@ from threading import (
 )
 
 from cv_bridge import CvBridge
-import cv2
 from geometry_msgs.msg import (
     Point,
     Pose,
@@ -266,7 +265,6 @@ class HL2SSROSBridge(Node):
                     "Warning! Image topic is not configured, so head pose data will not be published."
                 )
 
-
     def connect_hl2ss_pv(self) -> None:
         """
         Creates the HL2SS PV client and connects it to the server on the headset.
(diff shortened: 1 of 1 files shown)


ros/angel_system_nodes/angel_system_nodes/hl2ss_ros_bridge.py
```diff
@@ -270,6 +270,16 @@ def connect_hl2ss_pv(self) -> None:
...
 
+        # Get the camera parameters for this configuration
+        pv_cam_params = hl2ss.download_calibration_pv(
+            self.ip_addr,
+            self.pv_port,
+            self.pv_width,
+            self.pv_height,
+            self.pv_framerate
+        )
+        self.camera_intrinsics = [float(x) for x in pv_cam_params.intrinsics.flatten()]
```
Purg, line 281: Do we know if this is "correct"? I'm recalling we had that trouble when we switched from a deprecated MS plugin where the intrinsic of the camera became incorrect and we had to hard-code some matrix.

joshanderson-kw: No, unfortunately it doesn't match our hard-coded values. It's the values for focal length are -985 and 984, whereas our values are `FOCAL_LENGTH_X = 1.630`, `FOCAL_LENGTH_Y = 2.5084`.

I wonder if there is some unit conversion I am just not understanding.

Purg: I suppose let's keep the hard-coded intrinsics in our back pocket if we find that we are not able to use this intrinsics value correctly.

Purg (approved):
LGTM pending dependency merger.

joshanderson-kw:
> LGTM pending dependency merger

Conflicts should be fixed here now.

PR #216, opened 2023-04-19
Add Voice Activity Detection

derkmed:
Adds a new node for Voice Activity Detection (VAD) to reduce ASR processing latency. Previously, ASR was dependent on the `asr_req_segment_duration` parameter set to 8s. In order to reduce this latency, we propose a new upstream node that incrementally accumulates fixed-unit intervals (for example, 1s or 2s which are both much smaller than 8s) and runs voice activity detection on the accumulated audio. Once detected voice activity ends reasonably away from the end of the most recently appended fixed-unit, accumulated audio up to the detected voice activity end can be published to the subscribing ASR node.

This makes the VAD node the new bottleneck. However, its ability to process messages should have a lower latency.

See https://github.com/ColumbiaNLP/angel-system-speech-processor for the Flask Server with ASR + VAD functionality. Long-term, we will devise a plan on relocating this code onto the gpu-equipped host of the angel system.

### DIFF
--- a/ros/angel_system_nodes/angel_system_nodes/voice_activity_detector.py
+++ b/ros/angel_system_nodes/angel_system_nodes/voice_activity_detector.py
@@ -9,23 +9,41 @@ import wave
 from nltk.tokenize import sent_tokenize
 import rclpy
 from rclpy.node import Node
+from rclpy.parameter import Parameter
 import simpleaudio as sa
 
 from angel_msgs.msg import HeadsetAudioData, Utterance
 
+FLOAT32_BYTE_LENGTH = 4 # Number of Bytes in a float32.
 
 IN_AUDIO_TOPIC = "input_audio_topic"
 OUT_VOICE_ACTIVITY_TOPIC = "output_voice_activity_topic"
+
+# This parameter points to a server that returns a list of [start, end]
+# timestamp intervals of detected voice activity. Returned responses can be
+# emtpy if no voice activity was detected.
 VAD_SERVER_URL = "vad_server_url"
 
-#This parameter controls how frequently Voice Activity Detection should be
-# queried, e.g. every 1s of Input Audio should be accumulated for voice
-# activity detection.
-INTERVAL_UNIT_SEC_DURATION = "interval_unit_sec_duration"
+# This parameter controls how frequently Voice Activity Detection should be
+# queried. For example, a value of 1 specifies that each second's worth of audio
+# should be appended to an accumulating bytearray. The accumulated bytearray
+# should then be sent to the VAD_SERVER_URL to request voice activity detection.
+VAD_SEC_CADENCE = "vad_cadence"
 
-# TODO (derekahmed) We should figure out how this value was derived
-# and make this a constant accordingly.
-WAV_SAMPLE_WIDTH = 4  # Derived from audio_player.py
+# This parameter helps determine when accumulated audio that has detected vocal
+# inactivity should be published. If detected vocal inactivity begins
+# VAD_SEC_MARGIN seconds before the end of the "current" VAD_SEC_CADENCE,
+# then vocal activity is likely not ongoing / has ended. The vocal activity
+# should then be sliced from accumulated audio and published. Otherwise,
+# vocal activity may have been detected and is ongoing. Smaller values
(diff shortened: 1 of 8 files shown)


tmux/eval_asr_w_vad.yml
```diff
@@ -0,0 +1,76 @@
...
+
+# Specifies (by index) which pane of the specified window will be selected on project startup. If not set, the first pane is used.
+# startup_pane: 1
+
+# Controls whether the tmux session should be attached to automatically. Defaults to true.
+# attach: false
+
+windows:
+  # - ros_bag_play: ros2 bag play <<PATH_TO_BAG_FILE>>
+  - ros_bag_play: ros2 bag play /angel_workspace/ros_bags/rosbag2_2023_03_01-17_28_00/rosbag2_2023_03_01-17_28_00_0.db3
```
derkmed, line 57: Please replace the `ros_bag_play` window configurations prior to merging.

tmux/eval_vad.yml
```diff
@@ -0,0 +1,64 @@
...
+  - vad: ros2 run angel_system_nodes voice_activity_detector --ros-args
+      -r __ns:=${ROS_NAMESPACE}
+      -p input_audio_topic:=HeadsetAudioData
+      -p output_voice_activity_topic:=DetectedVoiceData
+      -p vad_server_url:=http://communication.cs.columbia.edu:55667/vad
+      -p vad_cadence:=2
+      -p vad_margin:=0.25
+      -p max_accumulation_length:=10
+  # - ros_bag_play: ros2 bag play <<PATH_TO_BAG_FILE>>
+  - ros_bag_play: ros2 bag play /angel_workspace/ros_bags/rosbag2_2023_03_01-17_28_00/rosbag2_2023_03_01-17_28_00_0.db3
```
derkmed, line 64: Please replace the `ros_bag_play` window configurations prior to merging. Also, please re-arrange the ordering to be aligned with current convention.

ros/angel_system_nodes/angel_system_nodes/voice_activity_detector.py
```diff
@@ -0,0 +1,368 @@
...
+        with self.audio_stream_lock:
+            audio_prefix = audio_data[:split_byte_idx]
+            self.log.info(f"Split @ {split_timestamp}s: ({split_byte_idx})")            
+            if self.debug_mode:
+                self.log.info("Writing VOCAL ACTIVITY DETECTED prefix of " +\
+                          "currently accumulated audio data to \"split\" " +\
+                          "temporary file.")
+                self._create_temp_audio_file(
+                    audio_prefix, sample_rate, num_channels,
+                    prefix=f"split-{self.split_counter}-")
```
derkmed, line 285: @Purg @joshanderson-kw 
We evaluated this on the ros bag that @Purg most recently provided, which contained ~30s of audio. However, the VAD .wav file splits created here (when debugging mode is turned on) are of a noticeably worse quality than the source audio .wav file that accompanied this ros bag. My understanding is this source .wav file was recorded and then placed in the ros bag file. This leads me to suspect possible downsampling of audio data placed in a ROS Bag. Does this observation align with either of your understandings/expectations? We want to make sure there is no unexpected bug that is causing this behavior.

_**Remarkably & most importantly, we noticed no drop in performance on ASR.**_

tmux/2023-03-demo-6-vad-asr-tool copy.yml
```diff
@@ -0,0 +1,151 @@
...
+      -p vad_server_url:=http://communication.cs.columbia.edu:55667/vad
+      -p vad_cadence:=2
+      -p vad_margin:=0.25
+      -p max_accumulation_length:=10
+  - asr: ros2 run angel_system_nodes asr --ros-args
+      -r __ns:=${ROS_NAMESPACE}
+      -p audio_topic:=DetectedVoiceData
+      -p utterances_topic:=UtterancesTopic
+      -p asr_server_url:=http://communication.cs.columbia.edu:55667/asr
+      -p asr_req_segment_duration:=4.0
```
derkmed, line 112: @ovenmitt this  is the new addition  for running Voice Activity Detection prior to ASR.

PR #217, opened 2023-04-19
README/Unity Tweaks

Purg:
I needed to update software packages on my Windows machine, leading me to make sure I could still build a package. I found some updates to the README and a few other things:
* Update to .Net API Level 4.x (defaulted to .Net 2.0 standard). I thought it might fix the following but no dice.
* Found that there were a few lines in CU scripts that were causing script compilation errors. I'm not sure if these are necessary, but I was able to export and build with them commented out. @ovenmitt please let me know if those include lines are indeed required and I can retract that commit. I'm just not sure how to proceed with my script compilation error with them there.

### DIFF
--- a/unity/ARUI/ProjectSettings/ProjectSettings.asset
+++ b/unity/ARUI/ProjectSettings/ProjectSettings.asset
@@ -705,7 +705,8 @@ PlayerSettings:
   gcIncremental: 1
   assemblyVersionValidation: 1
   gcWBarrierValidation: 0
-  apiCompatibilityLevelPerPlatform: {}
+  apiCompatibilityLevelPerPlatform:
+    WSA: 3
   m_RenderingPath: 1
   m_MobileRenderingPath: 1
   metroPackageName: Test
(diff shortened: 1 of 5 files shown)


ovenmitt:
@Purg you can delete those. Unity creates imports automatically (unfortunately), I will investigate why to avoid committing them in the future. - I do get a compilation error on my side as well.

unity/ARUI/ProjectSettings/ProjectSettings.asset
```diff
@@ -705,7 +705,8 @@ PlayerSettings:
   gcIncremental: 1
   assemblyVersionValidation: 1
   gcWBarrierValidation: 0
-  apiCompatibilityLevelPerPlatform: {}
+  apiCompatibilityLevelPerPlatform:
+    WSA: 3
```
joshanderson-kw, line 709: What is this settings change?

Purg: That is the change to use .Net API level 4.x (supposed to include more up to date APIs). Our original Unity/Visual Studio installation with the versions we recorded came with .Net 5.x (which is now outdated anyway in lieu of 6.x/7.x). I made this change initially to try to support the APIs now commented out, but didn't actually work, but seemed like a sensible thing to do given the mentioned context.

joshanderson-kw (approved):

PR #220, opened 2023-04-21
Task update message translation to BBN update message

Purg:
Add note to support translating our output TaskUpdate messages into the BBN update format (ROS2).

This required adding to the TaskUpdate message information about the timestamp of the latest sensor data that went into the task update estimation. Additional support was added to the HMM Task Monitor node to fill out this new field.


ros/bbn_integration_py/setup.py
```diff
@@ -0,0 +1,29 @@
...
+    install_requires=['setuptools'],
+    zip_safe=True,
+    maintainer='Paul Tunison',
+    maintainer_email='paul.tunison@kitware.com',
+    # ---
+    # Shared with properties in package.xml.
+    description='ROS2 nodes in python for BBN integration support',
+    license='BSD 3-Clause',
+    # ---
+    tests_require=['pytest'],
```
joshanderson-kw, line 23: When do these pytest scripts run? Are they run every time the package is built or just manually?

ros/bbn_integration_py/resource/bbn_integration_py
joshanderson-kw, line 1: What is the purpose of this file? Mostly just trying to understand

ros/bbn_integration_py/bbn_integration_py/nodes/task_to_bbn_update.py
```diff
@@ -0,0 +1,332 @@
...
+    """
+    Node to maintain task update state and translate to the BBN update format.
+
+    Each update output from the ANGEL system should correlate with one output
+    as BBN Update message.
+
+    Certain elements of the BBN Update output are currently hard-coded for demo
+    purposes:
+        * There is one casualty and its ID is 1
+        *
```
joshanderson-kw, line 42:
```suggestion
```

joshanderson-kw (commented):
Looks good! Just a few minor comments/questions.

ros/bbn_integration_py/setup.py
Purg: These are default values as populated by the `ros2 pkg create` tooling when specifying that it's going to be an ament_python package. Our other packages like `ros/angel_system_nodes` have this set here as well. This would be used in CI testing if that was set up (oh boy do we have issues to fix in that case).

ros/bbn_integration_py/resource/bbn_integration_py
Purg: Also an automatically generated file from `ros2 pkg create --build-type ament_python`. It is an example data-file to be included in the ROS2 python package. It's inclusion here is innocuous.

ros/bbn_integration_py/bbn_integration_py/nodes/task_to_bbn_update.py
Purg: I think this is more of an incomplete thought that I need to finish. Thanks for pointing it out!

joshanderson-kw (approved):

PR #221, opened 2023-04-24
Update spatial mapper node

joshanderson-kw:
- Replace trimesh library with open3d. The original trimesh ray casting function was taking too long (over 10 seconds) for one detection, but with open3d the ray casting is much faster and can keep up with the object detections rate.
- Update spatial mapper to work with hl2ss
- Create `SpatialMeshes` msg to store arrays of `SpatialMesh` messages.

Depends on #214 and #215.


Purg:
I assume that this PR also depends on #214?

joshanderson-kw:
> I assume that this PR also depends on #214?

Yeah, it depends on #214 and #215

PR #224, opened 2023-05-02
Add activity detection (classifier) node

joshanderson-kw:


Purg (commented):
I want to get this off of draft and get this stub in for @mattbrown11 to work with off of master. We can then also create configs that use this targeting master instead of working off of topic branches the whole time. I think that is a simpler mental model.

ros/angel_system_nodes/angel_system_nodes/berkeley_activity_detector.py
```diff
@@ -0,0 +1,126 @@
...
+
+        }
+
+        activity_det_msg = ActivityDetection()
+        activity_det_msg.header.stamp = self.get_clock().now().to_msg()
+        activity_det_msg.header.frame_id = "ActivityDetection"
+
+        # Set the activity det start/end frame time to this frame's source stamp
+        activity_det_msg.source_stamp_start_frame = msg.source_stamp
+        activity_det_msg.source_stamp_end_frame = msg.source_stamp
```
Purg, line 97: I found in the HMM  Task monitor that it asserts that the end time is strictly-greater than the start time. We'll need to update this to track the time of the previous detection message and provide that as the start time. For the first detection message received either we can add a start frame of node start-up time (initialized at construction) or not publish a classifier result for the first message received. I'm biased towards the first.

ros/angel_system_nodes/angel_system_nodes/berkeley_activity_detector.py
```diff
@@ -0,0 +1,126 @@
...
+
+        activity_det_msg.label_vec = list(label_conf_dict.keys())
+        activity_det_msg.conf_vec = list(label_conf_dict.values())
+
+        self._activity_publisher.publish(activity_det_msg)
+        log.info("Publish activity detection msg")
+
```
Purg, line 104: extraneous newline

Purg (approved):

PR #230, opened 2023-05-05
Add new tourniquet tasks.

mattbrown11:
Defines the task step and activity labels.


config/activity_labels/recipe_tourniquet.yaml
Purg, line 1: Nitpick: probably appropriate to prefix this with `medical_` and not `recipe_`.

PR #232, opened 2023-05-05
Jupyter passthrough

Purg:


Purg:
@mattbrown11 confirms this has worked just fine for his workflow with no modifications.

PR #240, opened 2023-05-09
Add conf threshold as parameter in berkeley tm

hdefazio:


hdefazio:
@Purg PTAL, this is an old MR but has a minor change that's worth landing

PR #243, opened 2023-05-09
Add gracefull CTRL-C response from C&C client

Purg:
Adding a graceful capture and shutdown of any running system configuration on error or user-interrupt of the C&C client.


joshanderson-kw (approved):
LGTM

PR #246, opened 2023-05-10
Allow model to run without contact information

hdefazio:
- Fixes issue where we couldn't output multiple hands of the same class
- Makes the code more robust to missing contact data (for when we try to run a model that doesn't have the contact head)

### DIFF
--- a/angel_system/berkeley/demo/predictor.py
+++ b/angel_system/berkeley/demo/predictor.py
@@ -462,7 +462,7 @@ class VisualizationDemo_add_smoothing(object):
         # Determine how many predictions we have left
         num_instances = len(labels)
         
-        if obj_obj_contact_classes is not None:
+        if using_contact:
             assert num_instances == len(obj_obj_contact_scores) == len(obj_obj_contact_classes) == len(obj_hand_contact_scores) == len(obj_hand_contact_classes)
         else:
             assert num_instances == len(boxes)
--- a/angel_system/berkeley/demo/predictor.py
+++ b/angel_system/berkeley/demo/predictor.py
@@ -242,7 +242,7 @@ class VisualizationDemo(object):
                 yield process_predictions(frame, self.predictor(frame))
 
 class VisualizationDemo_add_smoothing(object):
-    def __init__(self, cfg, last_time, draw_output=True, tracking=True, number_frames=None, fps=None, instance_mode=ColorMode.IMAGE, parallel=False):
+    def __init__(self, cfg, last_time, use_contact=True, draw_output=True, tracking=True, number_frames=None, fps=None, instance_mode=ColorMode.IMAGE, parallel=False):
         """
         Args:
             cfg (CfgNode):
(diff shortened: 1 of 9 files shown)


angel_system/berkeley/demo/predictor.py
```diff
@@ -432,31 +433,39 @@ def postprocess_instances(self,
...
+        if using_contact:
             contact_flag, Contact_infos, contact_hand_flag, Contact_hand_infos = util.find_contact(boxes, labels, obj_obj_contact_scores, obj_obj_contact_classes, obj_hand_contact_scores, obj_hand_contact_classes)
-
+        else:
+            contact_flag = Contact_infos = contact_hand_flag = Contact_hand_infos = None
         # Determine how many predictions we have left
         num_instances = len(labels)
-        assert num_instances == len(obj_obj_contact_scores) == len(obj_obj_contact_classes) == len(obj_hand_contact_scores) == len(obj_hand_contact_classes)
+        
+        if obj_obj_contact_classes is not None:
```
Purg, line 465: `using_contact` is fed by the same comparison, so could technically use that here instead of duplicating the logic.

angel_system/berkeley/utils/data/dataloaders/load_bbn_medical_data.py
```diff
@@ -65,28 +66,27 @@ def bbn_activity_data_loader(videos_dir, video, step_map=None,
...
     if add_before_finished_task:
         print('Adding before and finished')
         # before task
         sub_step_str = step_map['step 1'][0][0].lower().strip().strip('.').strip()
-        end = gt_activity[sub_step_str][0]['start'] # when the first step starts
-        gt_activity['Not started'.lower()] = [{ 'start': 0, 'end': end }]
-            
+        if sub_step_str in gt_activity.keys():
+            end = gt_activity[sub_step_str][0]['start'] # when the first step starts
+            gt_activity['Not started'.lower()] = [{ 'start': 0, 'end': end }]
```
Purg, line 81: For keys like this where there is no substitutions, why type out capitalization and then call `.lower()`? Why not just write the string in lower case? I imagine the repeated calls to `.lower()` is extraneous computational cost.

In addition, constant strings like `"not started"` and `"finished"` could be constant values defined more globally to be reused, reducing the risk of something being wonky because one of these strings is spelled wrong.

Purg: Would having an import-able `constants.py` file somewhere that has things like this be a short lift here?

Purg: Otherwise I'll just be fine with the local reduction in computations.

hdefazio: I've added it to the sprint board. But I'd rather have that be a separate MR so I don't break anything in this one

Purg: :+1:

PR #248, opened 2023-05-11
Update poetry version, add kwcoco/kwimage

joshanderson-kw:
Update poetry 1.5.1
  - With the older version, I was getting an error when trying to add packages. Seems to be a common error recently with an update to urllib removing the `strict` attribute. https://gitlab.com/tezos/tezos/-/merge_requests/8658
  
 ```
 root@thadion:/angel_workspace# poetry add --lock kwcoco
Skipping virtualenv creation, as specified in config file.

  AttributeError

  'HTTPResponse' object has no attribute 'strict'
 ``` 
  
Add kwcoco/kwimage dependencies


Purg:
Did you run this dep update from within the workspace container? I'm not seeing the same metadata-section-updating behavior when I do it in there.

Purg:
I also was not able to reproduce your above build error.

joshanderson-kw:
> I also was not able to reproduce your above build error.

Hmmm let me try adding the dep within the workspace and see if that helps.

joshanderson-kw:
@Purg, would be interesting to see if you also get this error. I got this after doing a clean checkout of master, running angel-docker-pull, angel-docker-build, and then trying to add a package within angel_workspace.

Purg:
Rebased on current master and the lock file has been updated with the output from poetry v1.5.1.

PR #249, opened 2023-05-11
Add augmentations to training

hdefazio:


Purg:
Well, I'm not sure why I cannot change the target branch. @hdefazio are you able to make this target master?

PR #253, opened 2023-05-19
Update to hmm core to support full covariance matrices.

mattbrown11:
Update to hmm core to support full covariance matrices.


Purg:
@mattbrown11 Is this branch still relevant?

PR #255, opened 2023-05-24
Initial simple CI workflow

Purg:


joshanderson-kw (commented):
Very nice! Probably a later thing but would it make sense to eventually have a target pytest coverage? Or a target for new code?

joshanderson-kw (approved):

PR #260, opened 2023-06-20
Fix style issues on branch that didn't have CI applied

Purg:


Purg:
Keeping @hdefazio in the loop. I will just merge this once CI passes.

PR #262, opened 2023-06-27
Bug fixes in ptg eval

hdefazio:
- Added option to insert interstitial, before, and finished steps into the ground truth (supports the tourniquet step-based work)
- Bug fix: there were cases in `get_time_wind_range` where the rounding of `ind1` and `ind2` caused them to be equal or `ind1 > ind2`. This meant these cases were not correctly applied to `valid` when using them to index

### DIFF
--- a/angel_system/data/common/discretize_data.py
+++ b/angel_system/data/common/discretize_data.py
@@ -1,4 +1,6 @@
 import logging
+import warnings
+
 from typing import Sequence
 from typing import Tuple
 
@@ -48,6 +50,11 @@ def get_time_wind_range(
     ind1 = max([ind1, 0])
     ind2 = min([ind2, len(time_windows)])
 
+    # assert ind2 >= ind1, "Invalid time window indexes"
+    if ind2 < ind1:
+        # Rounding issue
+        warnings.warn(f"ind1: {ind1} is greater than ind2: {ind2} for start time {start} and ebd time {end}")
+        ind1 = ind2
     return ind1, ind2
 
 
@@ -77,6 +84,7 @@ def discretize_data_to_windows(labels, gt, detections, time_window, uncertainty_
     # Split by time window
     # ============================
     # Get time ranges
+    
     assert (
         time_window > uncertainty_pad
     ), "Time window must be longer than the uncertainty pad"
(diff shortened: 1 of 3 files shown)


angel_system/data/common/discretize_data.py
```diff
@@ -48,6 +50,13 @@ def get_time_wind_range(
     ind1 = max([ind1, 0])
     ind2 = min([ind2, len(time_windows)])
 
+    # assert ind2 >= ind1, "Invalid time window indexes"
+    if ind2 < ind1:
+        # Rounding issue
+        warnings.warn(
+            f"ind1: {ind1} is greater than ind2: {ind2} for start time {start} and ebd time {end}"
```
Purg, line 57:
```suggestion
            f"ind1: {ind1} is greater than ind2: {ind2} for start time {start} and end time {end}"
```

angel_system/data/common/discretize_data.py
```diff
@@ -104,19 +114,27 @@ def discretize_data_to_windows(labels, gt, detections, time_window, uncertainty_
             time_windows,
         )
 
-        valid[ind1:ind2] = True
-        correct_label = detections["class_label"][i].strip().rstrip(".")
+        correct_label = detections["class_label"][i].lower().strip().strip(".").strip()
```
Purg, line 117: Why are there so many `strip` calls? Are there really class labels like `  foo  .  `? Could this not be replaced by a single `.strip(' .')`? (not sure that your class label would have newlines in there, but you could add that too if needed...)

There are more instances of this in here.

angel_system/data/common/load_data.py
```diff
@@ -141,18 +141,21 @@ def activities_from_dive_csv(filepath: str) -> List[Activity]:
...
+
     for row in df.iterrows():
         i, s = row
         a_id = int(s[0])
         frame, time = time_from_name(s[1])
+
         if a_id not in id_to_activity:
             id_to_activity[a_id] = Activity(
-                s[9].lower().strip(),  # class label
+                s[9].lower().strip().strip(".").strip(),  # class label
```
Purg, line 158: Same comment as above. There are more instances of this in here.

angel_system/data/common/load_data.py
```diff
@@ -174,14 +177,76 @@ def activities_from_dive_csv(filepath: str) -> List[Activity]:
...
         f"entries: {filepath}"
     )
+
     return list(id_to_activity.values())
 
 
+def add_inter_steps(
+    gt, min_start_time, max_end_time, add_inter_steps=True, add_before_after_steps=True
+):
+    first_activity = min(gt, key=lambda a: a.start)
```
Purg, line 193: Missing doc-string. I'm confused what this is handling.

angel_system/data/common/load_data.py
```diff
@@ -174,14 +177,76 @@ def activities_from_dive_csv(filepath: str) -> List[Activity]:
...
         f"Some activities in source CSV do not have corresponding end time "
         f"entries: {filepath}"
     )
+
     return list(id_to_activity.values())
 
 
+def add_inter_steps(
+    gt, min_start_time, max_end_time, add_inter_steps=True, add_before_after_steps=True
+):
```
Purg, line 192: Missing type annotations.

angel_system/data/common/load_data.py
```diff
@@ -174,14 +177,76 @@ def activities_from_dive_csv(filepath: str) -> List[Activity]:
...
     assert np.inf not in {a.end for a in id_to_activity.values()}, (
         f"Some activities in source CSV do not have corresponding end time "
         f"entries: {filepath}"
     )
+
     return list(id_to_activity.values())
 
 
+def add_inter_steps(
+    gt, min_start_time, max_end_time, add_inter_steps=True, add_before_after_steps=True
```
Purg, line 191: `ad_inter_steps` is both the name of the function *and* a parameter name. Consider an alternative parameter name.

angel_system/ptg_eval/activity_classification/evaluate_activity.py
```diff
@@ -36,15 +37,39 @@ def run_eval(args):
...
         if labels is None:
             labels = l_labels
         else:
             assert labels == l_labels, (
                 f"Subsequent gt/pred pair has disjoint label sets/orders. "
                 f"{labels} != {l_labels}"
             )
+        cleaned_labels = []
+        for label in labels:
+            cleaned_labels.append(label.lower().strip().strip(".").strip())
```
Purg, line 70: Why is this sprinkled all over the place?

angel_system/data/common/discretize_data.py
hdefazio: Yeah, there were a few instances of `foo  .  `. This is mainly trying to adjust for copy & paste spacing issues that we see a lot. I've at least set it up as a function so we can adjust how the strings are cleaned in one place. Would `.strip(" . ")` work instead?

hdefazio:
Also would like to close this MR in favor of #271 since it addresses most of these comments already

PR #267, opened 2023-07-11
Update ROS bridge to support Depth AHAT

yahskapar:
Draft of work to add RM Depth AHAT support back in for data collection. Currently, there are a few unknowns that may be best discovered by testing with the HoloLens 2 directly. As of writing this PR message, there is some trouble running ARUI with the HoloLens 2 at KRS and using it for testing without crashing.

### DIFF
--- a/ros/angel_system_nodes/angel_system_nodes/hl2ss_ros_bridge.py
+++ b/ros/angel_system_nodes/angel_system_nodes/hl2ss_ros_bridge.py
@@ -22,6 +22,8 @@ from angel_msgs.msg import (
     HeadsetAudioData,
     HeadsetPoseData,
     SpatialMesh,
+    # TODO: Is a message for depth AHAT needed here? Or from sensor_msgs?
+    # Currently re-using Image message
 )
 from angel_utils import declare_and_get_parameters, RateTracker
 from angel_utils.hand import JOINT_LIST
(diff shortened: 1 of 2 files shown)


ros/angel_system_nodes/angel_system_nodes/hl2ss_ros_bridge.py
```diff
@@ -172,6 +177,23 @@ def __init__(self):
             self._sm_thread = Thread(target=self.sm_publisher, name="publish_sm")
             self._sm_thread.daemon = True
             self._sm_thread.start()
+        if self._rm_depth_AHAT_topic != DISABLE_TOPIC_STR:
+            # Create frame publisher
+            self.ros_frame_publisher = self.create_publisher(
```
Purg, line 182: This shares the same name as the RGB image publisher, so this needs to be changed to be separate/unique.

Purg (approved):

ros/angel_system_nodes/angel_system_nodes/hl2ss_ros_bridge.py
```diff
@@ -22,6 +22,8 @@
     HeadsetAudioData,
     HeadsetPoseData,
     SpatialMesh,
+    # TODO: Is a message for depth AHAT needed here? Or from sensor_msgs?
+    # Currently re-using Image message
```
hdefazio, line 26: If this is addressed can we remove the comment?

Purg: A hint at this is retained by a lower comment pondering if other elements of the message received from HL2SS should be propagated down stream that we can determine at a later time.

PR #271, opened 2023-07-12
Add step evaluation to ptg eval

hdefazio:


hdefazio:
Depends on #262

angel_system/data/common/load_data.py
```diff
@@ -174,14 +187,292 @@ def activities_from_dive_csv(filepath: str) -> List[Activity]:
...
+        step_seq.append(
+            Step(
+                current_step_id,  # current id
+                sanitize_str(current_step),  # class_label
+                latest_sensor_ts,  # start
+                latest_sensor_ts,  # end
+                conf_vec[current_step_id],  # conf
+                completed_vec[current_step_id],  # completed
+            )
+        )
```
Purg, line 472: Not every message is a unique step, the output here is a streaming state message. I believe you will need to detect transitions if you want to have your output be discrete step time volumes.

Purg (commented):
I don't know if I'm specifically confident steps and activities can be scored the same way, but I suppose you are on the trajectory to trying it out and seeing what parts are applicable and make sense, and what doesn't.

angel_system/data/common/load_data.py
Purg: OBE

PR #277, opened 2023-08-16
Add object detection label configs for cooking

hdefazio:


config/object_labels/recipe_tea.yaml
```diff
@@ -0,0 +1,58 @@
+version: "1"
+title: "Tea"
+# Note: ids 5-20 are shared with coffee
```
joshanderson-kw, line 3: Does this mean that ids 5-20 in this file and the coffee files should match? They look different

joshanderson-kw (commented):
Where are these labels coming from? Is this just what we've been using for training?

Also, I see the notes about shared labels in several files, but the labels don't seem to match up.

config/object_labels/recipe_tea.yaml
hdefazio: Just updated the ids across all the files

hdefazio:
> Where are these labels coming from? Is this just what we've been using for training?

The coffee ones are the same ones that we've been using. The rest are new.

joshanderson-kw (approved):
LGTM!

PR #278, opened 2023-08-18
Fix bug where the knife wasn't clean between use

hdefazio:
This is intended to fix two issues that the annotation team found:

1. The knife was not cleaned in between use (so we were either slicing the banana with a dirty knife or scooping nutella with a dirty knife)
2. Sometimes "scoop nutella" would be immediately followed by "slice banana", an impossible combination.

### DIFF
--- a/angel_system/data/recipe_task_graph.py
+++ b/angel_system/data/recipe_task_graph.py
@@ -107,6 +107,13 @@ def print_recipe_order(recipe_title, labels, all_possible_orders, repeated_ids):
     # TODO: create sample bias towards regular order?
     recipe_order = random.choice(all_possible_orders)
 
+    if recipe_title == "Dessert Quesadilla":
+      # Remove extra clean
+      clean_1 = recipe_order.index(10)
+      clean_2 = recipe_order.index(11)
+
+      recipe_order.pop(min(clean_1, clean_2))
+
     print(f"\n{recipe_title}")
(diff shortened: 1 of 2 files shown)


hdefazio:
@Purg PTAL, this needs to land on Monday to unblock the annotators.

If you have a better solution, lmk. I wasn't able to work out a graph that put the "clean knife" step in the correct spot while also maintaining that we can start with either "slice banana" or "scoop nutella". So this fix is kinda hacky, but seems to produce okay results.

angel_system/data/recipe_task_graph.py
Purg, line 1: `black` is not happy with the spacing used in your additions (should be 4-space indentations to be consistent).

Purg:
It seems like we are encountering a real-world "critical resource" conflict (single knife for the recipe). Or alternatively considered an object is transformed such that another step is no longer satisfiable.

This solution is, for now, good enough as a short-term fix, but I do agree that we should come up with a more mature solution as this fix will not assist a next-gen task monitor algorithm.

Thinking out loud about the "real" issue: Either we will need to track "noun" resources as they are acquired/transformed by task steps, which can block other task steps, or consider that we have more than one instance of the nouns we are provided. I assume the latter is "easier" but may violate the recipe as provided, meaning that object resource usage tracking is the real way to go.

Purg (approved):

PR #287, opened 2023-09-14
Reorder the feature vectors

hdefazio:


hdefazio:
@cameron-a-johnson PTAL

PR #309, opened 2023-10-13
Add dummy multi task monitor node

joshanderson-kw:


joshanderson-kw:
@hdefazio, @Purg, can you check that my assumptions for handling multiple recipes are correct here?

-  Updated the `QueryTaskGraph` definition to return a list of `TaskGraphs` for the ARUI to use
-  `TaskUpdate` message stays the same (I'm thinking the ARUI can look at the `task_name` field to determine which recipe the update is for). 
- Add a multi recipe config for the multi task monitor node which is just a list of task config file paths for the node to monitor

joshanderson-kw:
@Purg / @hdefazio / @cameron-a-johnson, can I get a quick sanity check from someone on this? Tested this w/ Bettina and the new ARUI which handles multiple recipes.

ros/angel_system_nodes/angel_system_nodes/dummy_multi_task_monitor.py
Purg, line 1: I've recently organized this space more and there should now be a `task_monitoring/` sub-module this would be good to be included in instead of at the root level.

config/tasks/multi-task-config.yaml
```diff
@@ -0,0 +1,17 @@
...
+# Task title for display purposes.
+title: "Multi task example"
+
+# Layout of the steps that define this task.
+tasks:
+  # Item format:
+  # - id: Identifying integer for the task.
+  # - config_file: The task config file representing this task.
+  - id: 0
+    config_file: "./config/tasks/task_steps_config-recipe_coffee_trimmed_v3_allows_skip.yaml"
```
Purg, line 13: SInce these are relative paths, it should be noted what this is expected to be relative to, e.g. where the tool is run, `${ANGEL_WORKSPACE}`, etc.

ros/angel_msgs/srv/QueryTaskGraph.srv
```diff
@@ -1,6 +1,12 @@
-# Get the full task graph for the current task.
+# Get all of the task graphs for the current tasks.
+
+# This field is required b/c the ROS TCP bridge cannot handle deserializing an
```
Purg, line 3: I love your addition of descriptions! :100: Chesterton's fence will be harder to move.

tmux/demos/multi_task_monitor_demo.yml
```diff
@@ -0,0 +1,73 @@
...
+  - datahub: ros2 run ros_tcp_endpoint default_server_endpoint --ros-args
+      -r __ns:=${ROS_NAMESPACE}
+      -p ROS_IP:=0.0.0.0
+  # HL2SS connection
+  #- hl2ss_bridge: ros2 run angel_system_nodes hl2ss_ros_bridge --ros-args
+  #    -r __ns:=${ROS_NAMESPACE}
+  #    -p ip_addr:=${HL2_IP}
+  #    -p image_topic:=PVFramesBGR
+  #    -p hand_pose_topic:=HandJointPoseData
+  #    -p audio_topic:=HeadsetAudioData
```
Purg, line 57: Could probably remove this commented out block.

ros/angel_system_nodes/angel_system_nodes/dummy_multi_task_monitor.py
```diff
@@ -0,0 +1,360 @@
...
+        with open(self._config_file, "r") as f:
+            config = yaml.safe_load(f)
+
+        self._task_state_dict: Dict[int, TaskStateInformation] = {}
+        for t in config["tasks"]:
+            # Load the config for this task
+            with open(t["config_file"], "r") as f:
+                task_config = yaml.safe_load(f)
+
+            n_steps = task_config["steps"][-1]["id"]
```
Purg, line 104: Using the last ID might not be reliable. Would it not be better to take the `len(...)` here?

ros/angel_system_nodes/angel_system_nodes/dummy_multi_task_monitor.py
```diff
@@ -0,0 +1,360 @@
...
+        # Initialize ROS hooks
+        self._task_update_publisher = self.create_publisher(
+            TaskUpdate, self._task_state_topic, 1
+        )
+        self._task_graph_service = self.create_service(
+            QueryTaskGraph, self._query_task_graph_topic, self.query_task_graph_callback
+        )
+
+        # Store module for local use.
+        self._keyboard = keyboard
```
Purg, line 131: Why do we have to do this instead of just calling into the module normally?

ros/angel_system_nodes/angel_system_nodes/dummy_multi_task_monitor.py
```diff
@@ -0,0 +1,360 @@
...
+        """
+        Clear the error cache any IDs greater than the given.
+        """
+        s = np.asarray(list(self._steps_skipped_cache))
+        to_remove_ids = s[s > current_hmm_step_id]
+        if to_remove_ids.size:
+            self.get_logger().info(
+                f"Clearing step IDs from skipped cache: " f"{to_remove_ids.tolist()}"
+            )
+        [self._steps_skipped_cache.remove(_id) for _id in to_remove_ids]
```
Purg, line 149: Why do this instead of a normal for-loop? It seems like this would allocate a spurious list instance that is immediately destroyed.
```suggestion
        for _id in to_remove_ids:
            self._steps_skipped_cache.remove(_id)
```

ros/angel_system_nodes/angel_system_nodes/dummy_multi_task_monitor.py
```diff
@@ -0,0 +1,360 @@
...
+
+        message = TaskUpdate()
+
+        # Populate message header
+        message.header.stamp = self.get_clock().now().to_msg()
+        message.header.frame_id = "Task message"
+        message.task_name = task_state.task_title
+        if task_state.latest_act_classification_end_time is not None:
+            message.latest_sensor_input_time = nano_to_ros_time(
+                int(task_state.latest_act_classification_end_time * 1e9)
```
Purg, line 173: FWIW the `1e9` is defined as a constant in the same module you imported `nano_to_ros_time` from that is pre-integer-ized.

ros/angel_system_nodes/angel_system_nodes/dummy_multi_task_monitor.py
```diff
@@ -0,0 +1,360 @@
...
+    # sending duplicate notifications for the same error.
+    previous_step_skip = None
+    current_step_skip = None
+    # Track the indices of steps we have emitted skip errors for in order
+    # to not repeatedly emit many errors for the same steps.
+    # This may be removed from if step progress regresses.
+    steps_skipped_cache: Set[int] = field(default_factory=set)
+    # Track the latest activity classification end time sent to the HMM
+    # Time is in floating-point seconds up to nanosecond precision.
+    latest_act_classification_end_time = None
```
Purg, line 59: Maybe a silly question: why not just keep this in integer nanoseconds? It seems like there is a bunch of seconds<->nanoseconds conversion happening below. Storing as nanoseconds natively might be simpler in efficiency.

ros/angel_msgs/srv/QueryTaskGraph.srv
joshanderson-kw: Haha yeah, I actually tried to remove the `test` field and it broke everything. So I should have left that description the last time we saw that error.

ros/angel_system_nodes/angel_system_nodes/dummy_multi_task_monitor.py
joshanderson-kw: Good question. This was left over from what I copied from `task_monitor_v2`, which had the optional keyboard advance feature and a delayed import of `keyboard`. Since this module only supports the keyboard, I think it makes sense to remove the `self._keyboard` piece and use the module normally.

config/tasks/multi-task-config.yaml
joshanderson-kw: Note added

ros/angel_system_nodes/angel_system_nodes/dummy_multi_task_monitor.py
joshanderson-kw: This was how `task_monitor_v2` did it. I went ahead and made the change to store in ns here though.

Purg (approved):

Purg:
See this one built with pycocotools just fine... idk.

PR #310, opened 2023-10-17
Various util functions

hdefazio:


Purg:
@hdefazio Are these changes required for the (future) functionality in #310?

angel_system/activity_hmm/train_activity_classifier.py
Purg, line 1: Am I reading correctly that `compute_feats` no longer has versioned output? This seemed to be so in flux before, how are we handling model attribution to which version of this function was used? Or is that not applicable any more.

angel_system/activity_hmm/train_activity_classifier.py
```diff
@@ -147,13 +144,11 @@ def compute_feats(
...
 
-            try:
-                obj_obj_contact_state.append(ann["obj-obj_contact_state"])
-                obj_obj_contact_conf.append(ann["obj-obj_contact_conf"])
-                obj_hand_contact_state.append(ann["obj-hand_contact_state"])
-                obj_hand_contact_conf.append(ann["obj-hand_contact_conf"])
-            except KeyError:
-                pass
+        
+        #print(image_id)
```
Purg, line 148: Dead code.

angel_system/impls/detect_activities/detections_to_activities/utils.py
```diff
@@ -203,18 +199,21 @@ def obj_det2d_set_to_feature_by_method(
...
-    num_act = len(label_to_ind)
+    num_det_labels = len(label_to_ind)
     num_dets = len(label_vec)
 
-    act = np.zeros(num_act)
-    bboxes = [default_bbox for i in range(num_act)]
+    act = np.zeros(num_det_labels)
+    bboxes = [default_bbox for i in range(num_det_labels)]
+
+    #if label_vec == ['mug + tea bag + water', 'timer (off)', 'mug', 'thermometer (closed)', 'water jug lid', 'timer (off)', 'hand (right)', 'tea bag (dry)', 'water jug (open)', 'measuring cup']:
```
Purg, line 208: dead code

hdefazio:
> @hdefazio Are these changes required for the (future) functionality in #308 ?

Not #08, but this would delete the object/hand contact inputs, which would affect the activity classifier node and messages a little bit

Purg:
> > @hdefazio Are these changes required for the (future) functionality in #308 ?
> 
> Not #8, but this would delete the object/hand contact inputs, which would affect the activity classifier node and messages a little bit

Is the new model you referenced me towards going to use this form of vector?

angel_system/activity_hmm/train_activity_classifier.py
Purg: Self-answered during stand-up

config/activity_labels/all_recipe_labels.yaml
Purg, line 1: This file has a ton of repeated IDs. I'm assuming that will break something.

config/object_labels/recipe_oatmeal.yaml
```diff
@@ -46,6 +46,8 @@ labels:
     label: "measuring cup"
   - id: 25
     label: "container" # prep bowl
+  - id: 32
```
Purg, line 49: Why are these ID values jumping so high?

ros/angel_system_nodes/angel_system_nodes/yolov7_object_detector.py
Purg, line 1: This should be removed as the other branch, now merged, handled adding this node. This is not conflicting because my last branch reorganized node modules.

config/object_labels/recipe_oatmeal.yaml
hdefazio: We're trying to keep the id values unique across all recipes, so ids 26-31 are used somewhere else

config/activity_labels/all_recipe_labels.yaml
hdefazio: Just updated

hdefazio: Actually - moved all config updates to #315

PR #315, opened 2023-10-23
config updates

hdefazio:


hdefazio:
@cameron-a-johnson / @Purg

hdefazio:
@cameron-a-johnson  the configs under `config/tasks` are what you need. The activity ids are based on the ids in `config/activity_labels/all_recipe_labels.yaml`, which for now are different than each respective recipe activity config (I need to update the annotations before changing the ids in each file)

config/activity_labels/all_recipe_labels.yaml
```diff
@@ -10,216 +10,209 @@ labels:
   # - repeat: Optional field that indicated that the item must be repeated X times
```
Purg, line 10: We can probably remove the `depends` and `repeat` keys. They seem like they were for task step configs more thanctivity classification?

config/activity_labels/recipe_coffee+tea+dq+oatmeal.yaml
```diff
@@ -0,0 +1,173 @@
+version: "2.0"
+title: "Coffee + Tea + Dessert Quesadilla + Oatmeal"
+labels:
+  # Item:
+  # - id: Integer identification number of the label.
+  # - label: Short human-readable semantic string to succinctly represent the
+  #          label.
+  # - full_str: Complete sentence that fully represents the label.
+  # - depends: List of ids that the item depends on
```
Purg, line 9: Same here.

config/tasks/recipe_coffee.yaml
Purg, line 1: Are these recipes following the MIT-LL documented ones? If so, we should mention that here in a documenting comment. (Same for the other files)

config/activity_labels/all_recipe_labels.yaml
hdefazio: They were used to generate the out of order steps for the annotation team to create their data with. Idk if we'd want that on the tasks since they are higher level than the activities - we'd lose a lot of mixing of steps

Purg: There has just been a lot of concept mixing between activities and tasks/steps. I'm not clear on what the mixed legacy is any more.

hdefazio: Can I leave it for now until the tools that use the depends and repeat keys are updated?

config/tasks/recipe_coffee.yaml
hdefazio: added

PR #316, opened 2023-10-24
Add COCO output to TCN-HPL ROS2 node when appropriately optioned

Purg:


Purg:
PTAL @cameron-a-johnson @hdefazio

angel_system/tcn_hpl/predict.py
```diff
@@ -226,6 +229,88 @@ def predict(
...
+class ResultsCollector:
+    """
+    Utility class to collect TCN-HPL prediction results, encoding the ability
+    to output the results to a COCO file.
+
+    This is a beginning to generalizing results coco file aggregation and
+    output, based on output from `python-tpl/TCN_HPL/tcn_hpl/models/ptg_module.py`
+    in `on_validation_epoch_end` and `on_test_epoch_end` methods.
+
+    TODO: Promote into TCN_HPL repo?
```
hdefazio, line 254: This makes sense to me in the TCN_HPL repo, otherwise LGTM

Purg: I didn't put it there because I didn't see a direct path to making use of it within existing functionality there yet, as there would be debugging to do there, so its just used in the angel_system content right now.

PR #317, opened 2023-10-24
Add logging for tracing

Purg:
Added, or elevated to info level, logging to indicate when detector/classifier nodes received component data and perform processing.


Purg:
PTAL @cameron-a-johnson

PR #319, opened 2023-10-25
Update feedback generator to listen to SystemTestResponse topic

joshanderson-kw:
Update the feedback generator node to listen to `SystemTextResponse` messages and forward those to the ARUI as notifications (`AruiUserNotification`).


joshanderson-kw:
@derkmed, FYI I made a live system config here that I tested this with. Is #299 still needed?

derkmed:
> @derkmed, FYI I made a live system config here that I tested this with. Is #299 still needed?

Nope

Purg (approved):

PR #323, opened 2023-10-26
Dev/new step tracker

cameron-a-johnson:

PR #325, opened 2023-10-26
Add global step predictor ROS node

joshanderson-kw:
Depends on #323


Purg:
@joshanderson-kw I was assuming that I would not be looking to deep into this PR until #323 lands?

joshanderson-kw:
> @joshanderson-kw I was assuming that I would not be looking to deep into this PR until #323 lands?

Right

PR #326, opened 2023-10-26
Fix inputs to hl2ss_ros_bridge based on updates to that node

hdefazio:


hdefazio:
@cameron-a-johnson  / @Purg

tmux/record/record_ros_bag.yml
```diff
@@ -19,7 +19,7 @@ root: <%= ENV["ANGEL_WORKSPACE_DIR"] %>
 # on_project_start: command
 on_project_start: |
   export ROS_NAMESPACE=${ROS_NAMESPACE:-/kitware}
-  export HL2_IP=${HL2_IP:-192.168.1.101}
+  export HL2_IP=${HL2_IP:-192.168.1.4}
```
Purg, line 22: Why is the IP of the Hololens not '.101'?

hdefazio: Here at KRS it is .4 on the wifi.... I can keep the default .101 but maybe it's better for the default to be what we use on Remy?

Purg: sure, leave it at 4 then.

Purg (approved):

PR #327, opened 2023-10-27
Speed up TCN preprocessing and embedding

Purg:
Added memoization and vectorization to the largest bottlenecks.

For the same quantity of input processing we reduced this 
image

to this
image

which now shows that the proper DNN prediction as the bottleneck.


Purg:
PTAL @hdefazio @cameron-a-johnson

cameron-a-johnson (approved):

PR #352, opened 2023-11-01
fix to granular_to_broad converions

cameron-a-johnson:
An exciting fix!

The broad steps were being incremented from granular steps "a little late".

This is because an indexing tool I was using called "first_granular_step_per_broad_step" was actually the LAST granular step per broad step.

Example of the index:
```
 [0, 1, 3, 4, 6, 7, 8, 9, 11, 12, 14, 17, 18]
 ```
 granular step 3 = the LAST granular step of broad step 2 (so granular_step 2 is also part of broad_step 2). It was previously assumed to be the first granular step (so granular_step 2 was assumed to be part of step 1).
 
 We noticed latency in incrementing broad steps in the system. This will certainly improve that issue.

### DIFF
--- a/angel_system/global_step_prediction/global_step_predictor.py
+++ b/angel_system/global_step_prediction/global_step_predictor.py
@@ -290,7 +290,7 @@ class GlobalStepPredictor:
         granular_step_4
         """
         lgspbs = np.array(tracker["last_granular_step_per_broad_step"])
-        return len(np.nonzero(fgspbs < granular_step)[0])
+        return len(np.nonzero(lgspbs < granular_step)[0])
 
     def get_unique(self, activity_ids):
         """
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by cameron-a-johnson** [comment]
Some evidence that the mappings seem correct now:
image
The first `part_of_broad` was  equal to `1` before, which was wrong.

**Comment by cameron-a-johnson** [comment]
@hdefazio @joshanderson-kw PTAL.

File: angel_system/global_step_prediction/global_step_predictor.py [hunk 1]
```diff
@@ -303,7 +303,7 @@ def get_unique(self, activity_ids):
         indexes = np.unique(activity_ids, return_index=True)[1]
         return [activity_ids[index] for index in sorted(indexes)]
 
-    def get_first_granular_step_per_broad_step(self, steps):
+    def get_last_granular_step_per_broad_step(self, steps):
         """
         Get first substep index of each broad step.
```
Purg, line 308: Doc-string update?

cameron-a-johnson: Sure

cameron-a-johnson: Done. Thanks for pointing that out

joshanderson-kw (approved):

Purg (approved):

PR #362, opened 2023-11-02
Fix keyboard callback to properly publish messages for all tasks

hdefazio:


hdefazio:
CC: @cameron-a-johnson

PR #363, opened 2023-11-03
More granular task configs

hdefazio:


hdefazio:
Seems like half the steps are a bit late, the other half are on time, unsure if this is worth keeping

hdefazio:
rebase onto #364

PR #377, opened 2023-12-18
Reinstate bash code to generate xauth file for sharing with the container

Purg:
For some reason this was removed in an earlier commit relating to adding jupyter notebook runtime mounting with no indication as to why. Interpreting this as a mistake as I can confirm that X-pass through graphics tools no longer work with the current setup, but renew working when adding back in proper temporary xauth files for the containers to use.


hdefazio:
LGTM

PR #381, opened 2024-01-24
Moving around configs, Update medical configs

hdefazio:


hdefazio:
merging since failed test is unrelated

PR #393, opened 2024-04-15
Revert ubuntu back to 20.04 to fix ubuntu/python version issues

hdefazio:


hdefazio:
Getting the tests to pass will be in a separate MR.... but they are running now

pyproject.toml
```diff
@@ -88,6 +88,7 @@ tcn-hpl = {path = "python-tpl/TCN_HPL", develop = true}
 
 # Yolo v7 object detection
 yolov7 = {path = "python-tpl/yolov7", develop = true}
+click = "8.0.2"
```
Purg, line 91: What was motivating this requirement and pinned version regression?

Purg: I see, looked at the commit history and confirmed that black fails with the current version of installed click. I'm surprised that black's dependency listing allowed this. Good catch.

Purg: Actually, `black` is still not  running locally for me with this version of `click`.

hdefazio: I was seeing `ImportError: cannot import name '_unicodefun' from 'click'`. See related SO question: https://stackoverflow.com/questions/71673404/importerror-cannot-import-name-unicodefun-from-click

Purg: So, taking that into account, some modified feedback is
* let's move this specification down into the `dev-dependencies` section where `black` is specified (this could go right underneath the `black = "*"` line).
* Let's update the version specification for click to be `click = "*,<8.1"  # latest black for python 3.8.10 is not compatible with click >=8.1`

Purg (approved):

PR #394, opened 2024-04-15
Refactor feature vector generation

hdefazio:
- `time_from_name()` now requires the topic of the data it is being provided because the medical and cooking dataset images have different filename structures
        - this also required task specific versions of `time_from_name()` in their respective load_data files
- Added the ability to draw keypoints if they are present in `visualize_kwcoco_by_label()`
- Wrote hacky function `activity_label_fixes()` to map the BBN ground truth acitivty labels to our activity labels

MAIN POINT OF MR:
- Refactored `compute_feats()` in `train_actiivty_classifier.py` and `obj_det2d_set_to_feature()` and `obj_det2d_set_to_feature_by_method()` in `utils.py` so the joint distances would be added if the respective flags were turned on in `utils.py` and not always added in `train_activity_classifier.py`
        - this also resulted in refactoring the feature vector order to remove redundancies in the data and loops
- Added `plot_feature_vec()` to plot the various object/hand/joint points based on the feature vector and the kwcoco hand bounding boxes as a debug and visualization tool 
- Added `feature_version_to_options()` so we can access the corresponding feature vector flags in new spots - and thus updated `obj_det2d_set_to_feature()` to use this function


angel_system/data/medical/load_bbn_data.py
```diff
@@ -21,6 +22,29 @@
...
+    r"(?P<task>\w+)-(?P<vid>\d+)_(?P<frame>\d+).(?P<ext>\w+)"
+)
+
+
+def time_from_name(fname):
+    """
+    Extract the float timestamp from the filename.
+
+    :param fname: Filename of an image in the format
+        frame_<frame number>_<seconds>_<nanoseconds>.<extension>
```
cameron-a-johnson, line 36: Is there somewhere else that it would be useful to document this filename requirement?

cameron-a-johnson: Another idea: wherever such a filename requirement is assumed, we could add an "assert" checking for the right number of underscores

angel_system/data/common/load_data.py
```diff
@@ -29,6 +29,14 @@ def sanitize_str(str_: str):
     """
     return str_.lower().strip(" .")
 
+def time_from_name(fname, topic="cooking"):
```
hdefazio, line 32: add some assertion that the fname matches the RE_FILENAME_TIME parsing that we expect - if it doesn't, print RE_FILENAME_TIME

angel_system/data/common/load_data.py
```diff
@@ -29,6 +29,14 @@ def sanitize_str(str_: str):
...
+    if topic == "medical":
+        from angel_system.data.medical.load_bbn_data import time_from_name as tfn
+    elif topic == "cooking":
+        from angel_system.data.cooking.load_kitware_data import time_from_name as tfn
```
hdefazio, line 36: Is this the right way to handle multiple versions of this function? This was attempting to allow other functions to just import `from angel_system.data.common.load_data import time_from_name` without needing to recreate this if statement each time

angel_system/data/medical/load_bbn_data.py
cameron-a-johnson: ^From Hannah:

the id used to be parsed like this:
```
RE_FILENAME_TIME = re.compile(
    r"frame_(?P<frame>\d+)(?P<ts>\d+(?:|.)\d+).(?P<ext>\w+)"
)
```
So then you'd get a slightly-opaque error in this code:
```
    fname = os.path.basename(fname)
    match = RE_FILENAME_TIME.match(fname)
    time = match.group("ts")
```

Rough proposal: edit the function [here](https://github.com/PTG-Kitware/angel_system/blob/master/angel_system/data/common/load_data.py#L52) and always use it to acquire any arguments from the assumed filename above. Add some assertions (correct number of underscores, etc.) which try to enforce the above naming convention. If those assertions fail, the failure message should contain the `fname` convention that's documented in-line above.

@Purg , please tell me if you have better ideas or suggestions for the check I'm proposing here!

cameron-a-johnson: @hdefazio

angel_system/data/medical/load_bbn_data.py
```diff
@@ -207,42 +231,61 @@ def save_as_kwcoco(classes, data, save_fn="bbn-data.mscoco.json"):
...
-        label = "pull-remaining-strap"
-        label_id = 6
-    if activity_label == "secures" and target == "windlass":
-        label = "secure-strap"
-        label_id = 7
-    if activity_label == "writes_on" and target == "tourniquet_label":
-        label = "mark-time"
-        label_id = 8
-
+def activity_label_fixes(task, activity_label, target):
```
hdefazio, line 234: This is hacky but I can't think of a better way to map the given activity labels to our version of the label

cameron-a-johnson, line 234: This is hacky - the way BBN gave us activity GT, we parse it like this in angel_system/data/medical/load_bbn_data.py:
```
data = line.split("\t")

            # Find frame filenames
            start_frame = int(data[0])
            end_frame = int(data[1])

# Determine activity
            activity_str = data[2].strip().split(" ")
            hand = activity_str[0]
            activity = activity_str[1]
            target = activity_str[2] if len(activity_str) > 2 else None
```
Example output:
```
212     1128    hands      apply_pressure_to        casualty_wound
```
(Paraphrased)
"start frame"
"end frame"
"which hand is doing the activity" ("both" = "hands")
"activity" (generic, could apply to R18 or M2)
"target"

cameron-a-johnson: Should we add some text like this to the docstring for this function?

angel_system/data/medical/load_bbn_data.py
Purg: Late to the party. I would probably recommend updating the regex to have more specific captures for the expected pattern instead of capturing a variable pattern and then performing secondary parsing. Unless you're implying with the regex that the `_<nanoseconds>` portion is optional and sometimes not there. If the latter is the case, I would still recommend being explicit and having *two* regexs, one with and one without nanoseconds. Try matching against nanoseconds, if fail try parse against only seconds, if fail then raise an exception.
```python
RE_FILENAME_TIME_SEC = re.compile(
    r"frame_(?P<frame>\d+)_(?P<ts_sec>\d+)\.(?P<ext>\w+)"
)
RE_FILENAME_TIME_NANO = re.compile(
    r"frame_(?P<frame>\d+)_(?P<ts_sec>\d+)_(?P<ts_nano>\d+)\.(?P<ext>\w+)"
)
```

Purg (commented):
I'm not sure I am able to grasp what is going on at the larger picture view, but the smaller pieces that I could understand look good to me. See relatively minor comments.

I'll assume various leading/trailing whitespace things will be handled via a code formatting pass with `black`.

PR #395, opened 2024-04-17
Fix image resizing in object detector node, Add drawing joints to demo UI

hdefazio:
- Removes the optional resize in the object detector ros node
- Utilizes the built-in resizing from the yolo utilities
- Variable naming updates
- Moves the yolov8 detect function from yolov7 into the angel system
- Draw joints in the demo UI

### DIFF
--- a/ros/angel_utils/nodes/simple_2d_detection_overlay.cxx
+++ b/ros/angel_utils/nodes/simple_2d_detection_overlay.cxx
@@ -163,6 +163,7 @@ private:
   using frame_map_t = std::map< size_t, sensor_msgs::msg::Image::SharedPtr >;
 
   frame_map_t m_frame_map;
+  std::map< size_t, cv_bridge::CvImagePtr > m_drawn_frame_map;
 };
 
 // ----------------------------------------------------------------------------
@@ -347,9 +348,21 @@ Simple2dDetectionOverlay
     return;
   }
 
-  // Make a copy of the message image to plot on.
-  cv_bridge::CvImagePtr img_ptr =
-    cv_bridge::toCvCopy( find_it->second, "rgb8" );
+  // Check if we have already drawn something on this image
+  auto find_drawn_it = m_drawn_frame_map.find( source_nanosec_key );
+  bool insert = false;
+  cv_bridge::CvImagePtr img_ptr;
+  if( find_drawn_it == m_drawn_frame_map.end() )
+  {
+    insert = true;
+    // Make a copy of the message image to plot on.
+    img_ptr =
+      cv_bridge::toCvCopy( find_it->second, "rgb8" );
+  }
+  else
+  {
+    img_ptr = find_drawn_it->second;
+  }
 
   size_t num_detections = det_set->num_detections;
   size_t num_labels = det_set->label_vec.size();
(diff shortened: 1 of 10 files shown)


ros/angel_utils/nodes/simple_2d_detection_overlay.cxx
```diff
@@ -406,6 +423,129 @@ ::collect_detections( ObjectDetection2dSet::SharedPtr const det_set )
...
+      joints_msg->header.frame_id.c_str(),
+      find_it->second->header.frame_id.c_str()
+      );
+    return;
+  }
+
+  // Only plot the patient skeleton
+  if(joints_msg->hand != "patient")
+  {
+    return;
```
Purg, line 470: Maybe add a warning message for here as well. E.g. there should only be "patient" joints travelling across the topic we utilize for this, so I imagine that other joints being on this topic would be an issue we want to be warned about.

ros/angel_utils/nodes/simple_2d_detection_overlay.cxx
```diff
@@ -406,6 +423,129 @@ ::collect_detections( ObjectDetection2dSet::SharedPtr const det_set )
...
+    {"back", {"left_upper_leg", "right_upper_leg"}},
+    {"left_upper_leg", {"left_knee"}},
+    {"left_knee", {"left_lower_leg"}},
+    {"left_lower_leg", {"left_foot"}},
+    {"right_upper_leg", {"right_knee"}},
+    {"right_knee", {"right_lower_leg"}},
+    {"right_lower_leg", {"right_foot"}}
+  };
+
+  static auto const COLOR_PT = cv::Scalar{ 0, 255, 0 };
```
Purg, line 498: This could probably be defined at the top of the file with the other statics, but here is fine unless we want to reuse this or coordinate with other colors for other overlays.

ros/angel_utils/nodes/simple_2d_detection_overlay.cxx
```diff
@@ -406,6 +423,129 @@ ::collect_detections( ObjectDetection2dSet::SharedPtr const det_set )
...
+  {
+    return;
+  }
+
+  // Make a copy of the message image to plot on.
+  cv_bridge::CvImagePtr img_ptr =
+    cv_bridge::toCvCopy( find_it->second, "rgb8" );
+
+  std::map<std::string, std::vector<double>> joint_positions = {};
+  std::map<std::string, std::vector<std::string>> connection_lines = { // mapping of joint line connections
```
Purg, line 478: This could probably be a static thing as well.

ros/angel_utils/nodes/simple_2d_detection_overlay.cxx
```diff
@@ -406,6 +423,129 @@ ::collect_detections( ObjectDetection2dSet::SharedPtr const det_set )
...
+    std::vector<double> first_joint = joint_positions[joint_name];
+    cv::Point pt1 = { 
+      (int) round( first_joint[0] ),
+      (int) round( first_joint[1] ) 
+    };
+
+    for(auto const& connecting_joint : joint_connections)
+    {
+      std::vector<double> connecting_pt = joint_positions[connecting_joint];
+      cv::Point pt2 = {
```
Purg, line 530: We could probably define `pt2` outside of this for-loop, just under `pt1`'s definition, and then assign to the variable inside this nested for-loop. The compiler may be optimizing this for us, but constantly re-allocating the memory for a type is a non-zero cost that should be easy to eliminate here.

tmux/demos/medical/Kitware-R18.yml
```diff
@@ -64,21 +67,17 @@ windows:
...
-    - simple_2d_overlay: ros2 run angel_utils Simple2dDetectionOverlay --ros-args
-        -r __ns:=${ROS_NAMESPACE}
-        -p topic_input_images:=image_0
-        -p topic_input_det_2d:=ObjectDetections2d
-        -p topic_output_images:=pv_image_detections_2d
-        -p filter_top_k:=5
+    - object_detector:
+        layout: even-vertical
+        panes:
+            - objects_hands_detection: ros2 run angel_system_nodes object_hand_detector --ros-args
```
Purg, line 73: Nitpick: Could we call this 
```suggestion
            - objects_and_hands_detection: ros2 run angel_system_nodes object_hand_detector --ros-args 
```

I keep reading this incorrectly and think that this is only detecting hands.

Purg (approved):
Relatively small comments I think. When they're addressed I approve, so I'll mark that now. Feel free to merge after addressing.

PR #402, opened 2024-04-22
Script to convert a video to a ros bag

hdefazio:


ros/angel_utils/scripts/convert_video_to_ros_bag.py
```diff
@@ -0,0 +1,99 @@
...
+        type=str,
+        default="video123.mp4",
+        help=f"Path to a mp4 video",
+    )
+    parser.add_argument(
+        "--bag-folder",
+        type=str,
+        default="video123",
+        help=f"Path to the ros bag folder that will be created",
+    )
```
Purg, line 91: See above about adding an image output topic name parameter.

ros/angel_utils/scripts/convert_video_to_ros_bag.py
```diff
@@ -0,0 +1,99 @@
...
+        try:
+            bag_writer.write(
+                NAMESPACE + topic_name,
+                serialize_message(image_msg),
+                frame_ts.nanoseconds,
+            )
+        except Exception as err:
+            # Truncating the error message because it printed out the whole image_msg input
+            print("error", type(err), str(err)[:400])
+            return
```
Purg, line 70: Errors should return non-zero exit codes (this will return a 0 exit code). Use `exit(1)`, or some other number, to exit the interpreter with the given return code. Or let the exception be raised and leave the try/catch and return code handling to the `main()` function.

Purg (approved):

PR #413, opened 2024-05-15
Add pose_repeat_rate to activity classifier

hdefazio:


angel_system/activity_classification/tcn_hpl/predict.py
```diff
@@ -112,6 +112,17 @@ def normalize_detection_features(
     if that is not desired.
 
     :param det_feats: Object Detection features to be normalized.
+    :param feature_version: Version of the feature conversion approach.
+    :param top_k_objects: Number top confidence objects to use per label, defaults to 1
+    :param image_width: Integer pixel width of the image that object detections
+        were generated on.
+    :param image_height: Integer pixel height of the image that object
+        detections were generated on.
+    :param num_det_classes: Number of object detection classes (note: DOES include the ahnd labels but DOES NOT include the patient and user labels)
```
Purg, line 121:
```suggestion
    :param num_det_classes: Number of object detection classes (note: DOES include the hand labels but DOES NOT include the patient and user labels)
```

angel_system/data/common/kwcoco_utils.py
```diff
@@ -105,7 +106,17 @@ def add_activity_gt_to_kwcoco(topic, task, dset, activity_config_fn):
...
+                label = matching_gt.iloc[0]["class_label"] # default to the first gt
+
+                # Hacky temporary fix
+                # In the medical data, step 1 can cover the same frames as other steps,
+                # So we attempt to find the "hidden" step instead
+                if len(matching_gt) > 1:
+                    possible_labels = []
+                    for i, row in matching_gt.iterrows():
+                        possible_labels.append(row["class_label"])
+                    label = max(possible_labels)
```
Purg, line 118: Checking that this is still relevant and the comment is up to date?

ansible/roles/provision-files/vars/main.yml
```diff
@@ -122,15 +122,15 @@ girder_file_downloads:
...
+      sha512: 11c417480ef3b178cea5e9ba01878e2c1d2978129db4d8ac5d7eb0a1e6bdec0da74fc54f6bd69c8d04311809fc78503fad5c57fe45c5ee09f0f4f68197e34be1
       dest: "{{ stage_dirs.activity_classifier }}/r18_tcn.ckpt"
     - file_id: 6606b6e9aa5c8de3874c3f4a
       sha512: 3c84333390ee6b12327bb7e5debed37149c7f95cc437b16939f77b599d1a0b3b8c4f0995820b95973170b94df695494000a1c45fbf46632267c212e125fe58a3
       dest: "{{ stage_dirs.activity_classifier }}/r18_mapping.txt"
     # Global Step predictor model
-    - file_id: 663e4c5d687336214e7cdbff
-      sha512: 12b5bf950f64d9609182f795f7243edb6438b3919d920f0f527127660bb632abd5917ed149c28aa5ebcef7e7f6d8ed6edff39b27286eb0ce37a1ff7de2da4c77
+    - file_id: 66464bf9687336214e7cdeae
+      sha512: bc7884c258cb8704372dd69a3e7d999eaf70d716d271982077c7216ef27ab01219ef1e488f6f9669e11a7e8aa6ffb9d8e07d74edc47294f90cc9e6a723f5a822
```
Purg, line 133: Double check which models this refers to?

tmux/demos/medical/Kitware-R18.yml
```diff
@@ -105,6 +105,7 @@ windows:
       -p det_topic:=ObjectDetections2d
       -p object_net_checkpoint:=${MODEL_DIR}/object_detector/r18_det.pt
       -p hand_net_checkpoint:=${MODEL_DIR}/object_detector/hands_model.pt
+      -p cuda_device_id:=1
```
Purg, line 108: Let's keep the model device changes local because that is going to be machine-dependent.

Purg (commented):
Mostly LGTM. I think I want to push up a change to the other config files to be symmetric with the changes you've made to R18, at least for `window_leads_with_objects:=true`. Is the `deactivate_thresh_frame_count` param change applicable to all of the other tasks as well?

angel_system/activity_classification/tcn_hpl/predict.py
hdefazio: @cameron-a-johnson question for the `deactivate_thresh_frame_count` value - but I don't think we'll really know until we have good models to test with the other tasks

angel_system/data/common/kwcoco_utils.py
hdefazio: Updated wording a bit - I think still relevant, this is fixing the issue that we may get step 1 AND step x as gt for a singular frame. I still think it's hacky though because it assumes an integer label. This will break with non integer labels and if the problematic class is a higher step id.

ansible/roles/provision-files/vars/main.yml
hdefazio: Added a "Girder item ID" field to the training spreadsheet that will match the `file_id` here. The GSP will always be trained on the TCN model that is being downloaded (I'd consider it an error if it didn't)

PR #415, opened 2024-05-16
Black formatting

hdefazio:


angel_system/data/common/calibrate_camera_bboxes.py
Purg, line 1: I assume `black` did not create a whole new file :3

hdefazio: thank you lol

Purg (approved):

PR #421, opened 2024-05-21
Refactors all dialogue system nodes to leverage DialogueUtterance

derkmed:
This will allow nodes to be reordered or dropped easily as desired. Furthermore, other nodes can easily be added using the more modular DialogueUtterance msg.


ros/angel_system_nodes/angel_system_nodes/audio/asr.py
```diff
@@ -4,12 +4,13 @@
 import requests
 import soundfile
 import tempfile
+from termcolor import colored
```
Purg, line 7: I see that the `termcolor` package is incidentally in our container environment because of our current dependency on `detectron2`. However, since we are now using this directly, this should become a main dependency of our project. If our `pyproject.toml` file, we should add `termcolor = "^2"` to the `[tool.poetry.dependencies]` section. Then, while in the `./angel-workspace-shell.sh` environment, `poetry lock --no-update` should be run to appropriately update the `poetry.lock` file.

I can also just make this change and push it to your branch here if desired.

Purg (changes requested):

Purg (dismissed):

Purg (dismissed):

Purg (approved):

derkmed (commented):
LGTM

PR #423, opened 2024-05-21
Refactors Intent Detection node to leverage and emit DialogueUtterance

derkmed:


Purg (commented):
LGTM. I can approve when the dependent PRs are approved/merged.

PR #438, opened 2024-06-10
angel_system: CSV Dive file generation, data generation with correct step GT, re-training

periakiva:
1. CSV Dive file generation: done in the PTG Data generation stage. current method requires a Dive CSV format for the GT
2. using action GT labels was mapped to step GT with hard-coded logic; we now simply use the step GT
3. retraining of all tasks with correct GT, with new config files


angel_system/data/common/kwcoco_utils.py
```diff
@@ -144,7 +144,7 @@ def add_activity_gt_to_kwcoco(topic, task, dset, activity_config_fn):
             dset.imgs[gid]["activity_gt"] = activity_label
 
     # dset.fpath = dset.fpath.split(".")[0] + "_fixed.mscoco.json"
-    dset.dump(dset.fpath, newlines=True)
+    # dset.dump(dset.fpath, newlines=True)
```
Purg, line 147: Should these lines be removed?

angel_system/data/medical/load_bbn_data.py
```diff
@@ -230,38 +230,54 @@ def save_as_kwcoco(classes, data, save_fn="bbn-data.mscoco.json"):
...
-        if activity_label == "writes_on" and target == "tourniquet_label":
+        elif activity_label == "writes_on" and target == "tourniquet_label":
             label = "mark-time"
             label_id = 8
+
+    elif task == "m3":
+        pass
+
+    elif task == "m4":
+        pass
```
Purg, line 272: Nitpick: These don't *really* need to be here since there is no `else` statement that you're trying to skip. A comment would suffice to describe that there are no fixes for activity m3/m4 tasks.

angel_system/data/medical/load_bbn_data.py
```diff
@@ -317,26 +338,41 @@ def bbn_activity_txt_to_csv(task, root_dir, output_dir, label_version):
...
         if video_name in KNOWN_BAD_VIDEOS:
             continue
-
+        
         print(action_txt_fn)
-        with open(action_txt_fn) as action_f:
-            lines = action_f.readlines()
-
+        # with open(action_txt_fn) as action_f:
+        #     lines = action_f.readlines()
```
Purg, line 344: Dead-code.

angel_system/data/medical/load_bbn_data.py
```diff
@@ -317,26 +338,41 @@ def bbn_activity_txt_to_csv(task, root_dir, output_dir, label_version):
...
+
```
Purg, line 345: Looks like there are standard style issues in this file. `black` needs to be run, though I just pushed a fix to correctly ignore the ZED wrapper package recently added. Rebase/merge in master and then make sure you run `black .` in your python env or `./angel-workspace-shell.sh` environment.

angel_system/data/medical/load_bbn_data.py
```diff
@@ -317,26 +338,41 @@ def bbn_activity_txt_to_csv(task, root_dir, output_dir, label_version):
...
-            data = line.split("\t")
-
-            # Find frame filenames
-            start_frame = int(data[0])
-            end_frame = int(data[1])
-
+        
+        f = open(action_txt_fn, "r")
+        text = f.read()
+        f.close()
```
Purg, line 357: A more robust version of this is:
```suggestion
        with open(action_txt_fn, "r") as f:
            text = f.read()
```
We generally don't want to be calling `.close()` on files and should leave that up to a context manager in case there is any errors during operation so that resource cleanup happens in due course.

angel_system/data/medical/load_bbn_data.py
```diff
@@ -345,15 +381,15 @@ def bbn_activity_txt_to_csv(task, root_dir, output_dir, label_version):
...
 
             # Determine activity
-            activity_str = data[2].strip().split(" ")
-            hand = activity_str[0]
-            activity = activity_str[1]
-            target = activity_str[2] if len(activity_str) > 2 else None
+            # activity_str = data[2].strip().split(" ")
+            # hand = activity_str[0]
+            # activity = activity_str[1]
+            # target = activity_str[2] if len(activity_str) > 2 else None
```
Purg, line 387: Dead code, more below.

angel_system/data/medical/load_bbn_data.py
```diff
@@ -365,7 +401,8 @@ def bbn_activity_txt_to_csv(task, root_dir, output_dir, label_version):
 
                 track_id += 1
         csv_f.close()
-
+        
+    return used_videos
```
Purg, line 405: This looks like a new return statement for this function. Please update the doc-string accordingly.

angel_system/data/common/kwcoco_utils.py
periakiva: Yes - rewriting over our data makes no sense (this function does that).

angel_system/data/medical/load_bbn_data.py
Purg: Did you mean to remove the `task == "m5"` condition that was here in your edit previously?

periakiva: Yes - we don't use this function for any of the tasks anymore. no need to map actions to steps if we already have the step ground truth (not sure why it was written to begin with).

angel_system/data/medical/load_bbn_data.py
```diff
@@ -299,11 +301,18 @@ def bbn_activity_txt_to_csv(task, root_dir, output_dir, label_version):
...
+    
+    :return: A list containing names of used videos
 
     """
-    print(f"{root_dir}/*/*_action_labels_by_frame.txt")
-
-    action_fns = glob.glob(f"{root_dir}/*/*.action_labels_by_frame.txt")
+    print(f"{root_dir}/*/*_skill_labels_by_frame.txt")
+
+    # action_fns = glob.glob(f"{root_dir}/*/*.action_labels_by_frame.txt")
```
cameron-a-johnson, line 310: Delete this line?

cameron-a-johnson:
@periakiva - Note that `code-style-py` failed. Could you run "black ." from the repo to fix those style issues?

Purg:
If we want to, this branch can now include the submodule reference update, otherwise I can do that separately as a branch off of master.

PR #442, opened 2024-06-12
Actually use the pose bounding box confidence threshold. For now use …

cameron-a-johnson:
…the default of 0.25.


cameron-a-johnson:
Also, I provisioned the latest/"greatest" R18 models here.

Now, instead of a strong tendency for the tracker to jump way ahead, this tracker much more frequently fails to progress, especially on steps 2 & 3 for my local testing.

Purg (dismissed):

Purg:
Style check is not happy.

Purg (approved):

PR #443, opened 2024-06-15
Delay simple overlay output

Purg:
Convert logic of the node to use a hard latency delay to perform drawing instead of previous logic.


Purg:
PTAL @cameron-a-johnson

PR #445, opened 2024-06-17
Dev/rst doc qa

cameron-a-johnson:


docs/system/angel_system_plus_speech_QA.rst
```diff
@@ -0,0 +1,79 @@
+Running Angel System with Speech QA
+======================
```
Purg, line 2: For RST files, these section demoninators are going to want to be the same character length as the section title. This will be a warning that is output by a Sphinx build. This comment applies to various sections below this.

docs/system/angel_system_plus_speech_QA.rst
```diff
@@ -0,0 +1,79 @@
...
+
+.. code-block:: bash
+
+   sudo apt update && apt install -y sox ffmpeg
+
+Running the Server
+----------------------
+
+Create conda environment
+----------------------
```
Purg, line 45: Was it intended for the "Create conda environment" section to be a child of "Running the Server"? If so, I recommend following the section "tiering" documentation [here](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections). Following that documentation, this section should use `^` characters.

docs/system/angel_system_plus_speech_QA.rst
```diff
@@ -0,0 +1,79 @@
...
+.. code-block:: bash
+
+   conda env create -f speech_server.yml
+   conda activate speech_server
+
+The server can then be instantiated with:
+
+.. code-block:: bash
+
+   export CUDA_VISIBLE_DEVICES=4; python speech_server.py
```
Purg, line 56: Theoretically, this should be equivalent to running `CUDA_VISIBLE_DEVICES=4 python speech_server.py`, but if what you have works, it works.

Purg:
See comments above still, if this was not already on your plate.

PR #449, opened 2024-09-27
June demo

Purg:
Re-created PR to re-open CI processing and provide a patch to ZED python package version.


cameron-a-johnson:
Re:

> "Docker image is currently too big for the default Github CI Runners... Zed bloat is mostly to blame"


Would it be viable to change the default Github CI Runners' maximum allowed docker image size? Especially since you mentioned hesitation to remove the Zed bloat for now.

Purg:
> Would it be viable to change the default Github CI Runners' maximum allowed docker image size? Especially since you mentioned hesitation to remove the Zed bloat for now.

I don't believe that is possible without paying Github money. l found something related to getting extremely nitty-gritty with Github automation where we would `rm ...` a bunch of non-critical stuff from the instance image host before proceeding into docker stuff, but it's a bit fragile and I don't think it would give us much room.

In the future, based on a convo with Brian, we may not continue to use the Zed stuff, so that may be removable. I might put up a Draft PR later today doing just that to set the stage for when it's appropriate.

cameron-a-johnson (approved):

PR #450, opened 2024-10-09
Minor updates

Purg:
These were some recent minor updates I had found in reintroducing myself into various parts of the codebase.

### DIFF
--- a/docker/workspace-build/Dockerfile
+++ b/docker/workspace-build/Dockerfile
@@ -4,7 +4,7 @@
 # Assumptions:
 # - ROS_DISTRO is defined with the name of a distribution installed under /opt/ros/
 #
-ARG ROS_VERSION
+ARG ROS_VERSION="foxy"
 FROM osrf/ros:${ROS_VERSION}-desktop AS base
 
 SHELL ["/bin/bash", "-c"]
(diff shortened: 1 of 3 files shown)


docker/workspace-build/Dockerfile
```diff
@@ -4,7 +4,7 @@
 # Assumptions:
 # - ROS_DISTRO is defined with the name of a distribution installed under /opt/ros/
 #
-ARG ROS_VERSION
+ARG ROS_VERSION="foxy"
```
cameron-a-johnson, line 7: Help me recall: what are the reasons we cannot upgrade from foxy? mmdet was one; were there others?

cameron-a-johnson: I was wondering if we should revisit this

cameron-a-johnson (approved):

docker/workspace-build/Dockerfile
Purg: mmdet/mmpose specifically.

cameron-a-johnson: Right. Noted, thanks.

josephvanpeltkw:
@Purg - this looks good to me!

PR #451, opened 2024-10-09
Remove ZED SDK Dependencies

Purg:
This brings our docker image size down from 24.7GB to 9.75GB.


josephvanpeltkw (approved):
Looks like we just need to handle the conflicts

PR #453, opened 2024-10-15
Dev/tensorrt hands

josephvanpeltkw:
Convert the hands detection model to tensorRT form


angel_system/object_detection/yolov8_detect.py
```diff
@@ -9,7 +9,7 @@ def predict_hands(hand_model: YOLOv8, img0: np.array, device: str, imgsz: int) -
     """
     width, height = img0.shape[:2]
     hands_preds = hand_model.predict(
-        source=img0, conf=0.1, imgsz=imgsz, device=device, verbose=False
+        source=img0, conf=0.1, device=device, verbose=False
```
Purg, line 12: Why remove the `imgsz` parameter? I vaguely recall this was pretty important so that we are inferencing with the same resolution that the model was trained on (768 or something).

pyproject.toml
```diff
@@ -48,7 +48,10 @@ mmcv = "==1.3.8"
 torch = {url = "https://download.pytorch.org/whl/cu111/torch-1.9.1%2Bcu111-cp38-cp38-linux_x86_64.whl"}
 torchvision = {url = "https://download.pytorch.org/whl/cu111/torchvision-0.10.1%2Bcu111-cp38-cp38-linux_x86_64.whl"}
 detectron2 = {url = "https://dl.fbaipublicfiles.com/detectron2/wheels/cu111/torch1.9/detectron2-0.6%2Bcu111-cp38-cp38-linux_x86_64.whl"}
-
+tensorrt = "==10.1.0"
+tensorrt-cu12 = "==10.1.0"
```
Purg, line 52: Do you need to specify these specifically, or are they effectively dependencies of `tensorrt`?

poetry.lock
Purg, line 1: A looooot of versions are updating here. How did you go about updating/adding the ultralytics and tensorrt dependencies? There are a few ways to add/update deps with poetry. Some will only add update what you ask for, and others will update *everything*. It seems like a route that updated *everything* was taken, which we want to avoid for system stability.

pyproject.toml
josephvanpeltkw: @Purg , when I didn't specify them they ended up being a different version and led to conflicts.

poetry.lock
josephvanpeltkw: @Purg , agreed - I'll redo this and get the smaller change version

angel_system/object_detection/yolov8_detect.py
josephvanpeltkw: @Purg , when this parameter is included, the model won't return any predictions (with this version of YOLOv8 which was needed to convert the model to tensorRT).  By removing this parameter, I started getting detections.  I'm looking into if there was a bug with this version.

Purg: One of the things I observed is that, with at least TensorRT conversion amongst possibly other versions, the image size is "baked in" to the model as part of the conversion. It may be the case that we need to convert the model with the appropriate image size we would provide here. Did you already do that when you converted the model to TensorRT and are still not getting outputs? If so, we may want to detect / indicate that the model is TensorRT as opposed to not, because this parameter is pretty required for correct predictions if we don't use the TensorRT converted model, and the current form of the node is not asserting/requiring a TensorRT input.

josephvanpeltkw: @Purg , the re-export with the `imgsz` parameter worked!

Purg (approved):

PR #454, opened 2024-10-16
Some high level helper scripts for BBN data extraction and ffprobe calls

Purg:


scripts/extract_bbn_video_ffprobe.bash
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
josephvanpeltkw, line 1: @Purg , could you add a comment with a quick description and possibly usage?

josephvanpeltkw (approved):
Looks ready to merge!

PR #460, opened 2024-10-23
Dev/adjust tcn coco saving

josephvanpeltkw:
Adjusts the saving of the coco file from the TCN ros node so that the annotations are in the annotation section instead of in the images


angel_system/activity_classification/tcn_hpl/predict.py
```diff
@@ -464,19 +464,30 @@ def collect(
...
-                packet["activity_gt"] = activity_gt
-            self._dset.add_image(**packet)
+                img["activity_gt"] = activity_gt
+            # save the gid from the image to link to the annot
+            gid = self._dset.add_image(**img)
+
+            # additional items to save
+            add_items = dict(
+                prob=list(activity_conf_vec),
+            )
```
Purg, line 486: Why do this instead of just adding a kwarg to the `_dset.add_annotation` method?
We can also just add the `score` field while we're here since that should be straightforward.
```python
            self._dset.add_annotation(
                image_id=gid, category_id=activity_pred,
                score=activity_conf_vec[activity_pred],
                prob=list(activity_conf_vec),
            )
```

josephvanpeltkw: agreed - I made that change

angel_system/activity_classification/tcn_hpl/predict.py
```diff
@@ -464,19 +464,29 @@ def collect(
...
             if name is not None:
-                packet["name"] = name
+                img["name"] = name
             if file_name is not None:
-                packet["file_name"] = file_name
+                img["file_name"] = file_name
             if activity_gt is not None:
-                packet["activity_gt"] = activity_gt
-            self._dset.add_image(**packet)
+                img["activity_gt"] = activity_gt
```
Purg, line 479: I don't think we should still be adding this non-standard field to the image objects. As long as we ensure there are equivalent image entries and that the added annotations reference their relevant images, the output COCO file will be separately interpretable to the ground truth file that was input.

ros/angel_system_nodes/angel_system_nodes/activity_classification/activity_classifier_tcn.py
```diff
@@ -649,12 +660,29 @@ def rt_loop(self):
...
                         self._activity_publisher.publish(act_msg)
                     except NoActivityClassification:
+                        # collect the results if we are saving to coco file
+                        if self._results_collector:
+                            # Prepare output message
+                            activity_msg = ActivityDetection()
+                            # set the only needed items for collection
+                            activity_msg.source_stamp_end_frame = window.frames[-1][0]
+                            activity_msg.conf_vec = [0.0 for x in self._model.classes]
+                            self._collect_results(activity_msg)
```
Purg, line 670: Why collect a zero-vector prediction for such a "failed" frame? I would think this is implied if there is an image in the output COCO file but no prediction. I am seeing that an image object was being added as part of the `collect` method, but I think I am realizing that is maybe more a historical artifact than a thing we should really be doing. Maybe the `ResultsCollector` class could be given an `add_image` method that we invoke before this try-catch to add the image metadata, and then only optionally collect the results annotation metadata on successful classification for that frame. Thoughts?

angel_system/activity_classification/tcn_hpl/predict.py
josephvanpeltkw: agreed - I'll drop this.

ros/angel_system_nodes/angel_system_nodes/activity_classification/activity_classifier_tcn.py
josephvanpeltkw: I agree - I think that sounds like a better approach

ros/angel_utils/python/angel_utils/activity_classification.py
```diff
@@ -192,13 +192,17 @@ def queue_image(
         # self.get_logger_fn().info(f"self.frames[-1][0] header stamp: {self.frames[-1][0]}")
         with self.__state_lock:
             # before the current lead frame?
-            if self.frames and time_to_int(img_header_stamp) <= time_to_int(
-                self.frames[-1][0]
+            if (
+                self.frames
+                and self.frames[-1][2] == image_frame_number
+                and time_to_int(img_header_stamp) <= time_to_int(self.frames[-1][0])
```
Purg, line 198: Were we getting situations where different frames had the same timestamp?

josephvanpeltkw: Yes, although I'm not sure if this is only an artifact of playing a bag that was created with our script.

josephvanpeltkw: Looks like this was caused by our script that converted videos to bags - I've got a fix in for that now, so I've backed this change out, @Purg

ros/angel_system_nodes/angel_system_nodes/activity_classification/activity_classifier_tcn.py
```diff
@@ -605,6 +615,7 @@ def rt_loop(self):
                     # out older data at and before the first item in the window.
                     self._buffer.clear_before(time_to_int(window.frames[1][0]))
 
+                    # set this to None to signal if we saved the image or not
```
Purg, line 618: This comment is not attached to anything any more.
```suggestion
```

ros/angel_system_nodes/angel_system_nodes/activity_classification/activity_classifier_tcn.py
```diff
@@ -762,7 +781,7 @@ def _process_window(self, window: InputWindow) -> ActivityDetection:
 
         return activity_msg
 
-    def _collect_results(self, msg: ActivityDetection):
+    def _collect_image(self, end_frame_time) -> int:
```
Purg, line 784: missing time type annotation
```suggestion
    def _collect_image(self, end_frame_time: Time) -> int:
```

Purg (approved):

PR #461, opened 2024-10-25
Integrate new TCN dataset and prediction

Purg:
Update to ROS side of things to make use of the extensive updated to the TCN submodule.


TRAIN_AND_RUN_README.md
```diff
@@ -1,61 +1,132 @@
...
+Consult a team member for login information.
+
+Each task ("skill") has their own sub-page ("In-Lab Data" link) from which sets
+of data are described and referred to.
+Data is stored on their SFTP server, to which the "Click to Download" links
+refer to.
+
+Storage of downloaded ZIP archives, and their subsequent extractions, should
+follow the pattern.
+A script is provided
```
josephvanpeltkw, line 75: recommend listing the location of the script for easy reference

tmux/demos/medical/Kitware-M2.yml
```diff
@@ -115,9 +118,9 @@ windows:
       -p image_ts_topic:=PVFramesBGR_ts
       -p det_topic:=ObjectDetections2d
       -p pose_topic:=PatientPose
+      -p activity_config_file:=${CONFIG_DIR}/activity_labels/medical/m2.yaml
       -p model_weights:=${MODEL_DIR}/activity_classifier/m2_tcn.ckpt
-      -p model_mapping:=${MODEL_DIR}/activity_classifier/m2_mapping.txt
-      -p model_det_label_mapping:=${ANGEL_WORKSPACE_DIR}/config/object_labels/medical/m2.json
+      -p model_config:=${MODEL_DIR}/activity_classifier/m2_config.yaml
```
josephvanpeltkw, line 123: Should we make these changes on the other medical yml files?

josephvanpeltkw (approved):

tmux/demos/medical/Kitware-M2.yml
Purg: I don't have any other models/configs for the other tasks, so it seemed to me that changing other configs would just be notional change. Maybe an alternative here is that don't just modify that existing config but create a new one. We would not be able to run the other configurations due to this API change, so I could move them to an "outdated" folder or something?

TRAIN_AND_RUN_README.md
Purg: I see I forgot to finish that sentence.

PR #471, opened 2024-11-12
Dev/camera metadata

josephvanpeltkw:
Convert the image_timestamp_relay to an image metadata topic.  Also does the same for the hl2ss node.  Adjusts down-stream processes like the latency node and activity_classifier_tcn node to consume the changed topic.  Drops the need for parameters into the tcn node for image width and height.

### DIFF
--- a/ros/angel_system_nodes/angel_system_nodes/activity_classification/activity_classifier_tcn.py
+++ b/ros/angel_system_nodes/angel_system_nodes/activity_classification/activity_classifier_tcn.py
@@ -164,8 +164,6 @@ class ActivityClassifierTCN(Node):
 
         self._act_topic = param_values[PARAM_ACT_TOPIC]
         self._act_config = load_activity_label_set(param_values[PARAM_ACT_CONFIG_FILE])
-        self._img_pix_width = 1280  # default size - replaced by metadata topic
-        self._img_pix_height = 720
         self._enable_trace_logging = param_values[PARAM_TIME_TRACE_LOGGING]
 
         self._window_lead_with_objects = param_values[PARAM_WINDOW_LEADS_WITH_OBJECTS]
@@ -412,14 +410,10 @@ class ActivityClassifierTCN(Node):
         """
         Capture a detection source image timestamp message.
         """
-        # set the image width and height
-        self._img_pix_width = msg.width
-        self._img_pix_height = msg.height
-
         log = self.get_logger()
         self._current_frame_number += 1
         if self.rt_alive() and self._buffer.queue_image(
-            None, msg.header.stamp, self._current_frame_number
+            None, msg, self._current_frame_number
         ):
             if self._enable_trace_logging:
                 log.info(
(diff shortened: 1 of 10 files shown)


ros/angel_system_nodes/angel_system_nodes/activity_classification/activity_classifier_tcn.py
```diff
@@ -161,16 +156,16 @@ def __init__(self):
...
         self._det_topic = param_values[PARAM_DET_TOPIC]
 
         self._pose_topic = param_values[PARAM_POSE_TOPIC]
         self._pose_repeat_rate = param_values[PARAM_POSE_REPEAT_RATE]
 
         self._act_topic = param_values[PARAM_ACT_TOPIC]
         self._act_config = load_activity_label_set(param_values[PARAM_ACT_CONFIG_FILE])
-        self._img_pix_width = int(param_values[PARAM_IMAGE_PIX_WIDTH])
-        self._img_pix_height = int(param_values[PARAM_IMAGE_PIX_HEIGHT])
+        self._img_pix_width = 1280  # default size - replaced by metadata topic
```
cameron-a-johnson, line 167: I hesitate to have a default value here that could lead to sneaky issues when we assume downstream the metadata topic has replace these width & height values with those of the video frame

cameron-a-johnson (dismissed):

Purg:
Build of the container in CI is failing. Please check in on that with a local build.

cameron-a-johnson:
> Build of the container in CI is failing. Please check in on that with a local build.

Oof I'm seeing "No space lef on device" in the build error logs:
```
/angel_workspace/ros/rosbag2/rosbag2_tests/test/rosbag2_tests/test_rosbag2_record_end_to_end.cpp:712:1: fatal error: error writing to /tmp/cckKfwtH.s: No space left on device
```

ros/angel_system_nodes/angel_system_nodes/data_publishers/hl2ss_ros_bridge.py
```diff
@@ -122,8 +121,8 @@ def __init__(self):
             self.ros_frame_publisher = self.create_publisher(
                 Image, self._image_topic, 1
             )
-            self.ros_frame_ts_publisher = self.create_publisher(
-                Time, self._image_ts_topic, 1
+            self.ros_frame_md_publisher = self.create_publisher(
```
Purg, line 124: There is a world where this is a service instead of a constantly published message, however this is definitely fine for now and is not going to exceed our bandwidth budgets.

tmux/demos/medical/old/Kitware-M3.yml
Purg, line 1: Why this "old" M3 file? If we're going to "modernize" this one, might as well bring a copy of it out of "old" and into the primary directory.

ros/angel_system_nodes/angel_system_nodes/activity_classification/activity_classifier_tcn.py
Purg: I would second this. It should probably be the case that the windowing functionality is updated to store `CameraInfo` instances instead of `Time` so that the window slices come with these metadata-carrying objects to assign image sizes to `FrameData` on a per-frame basis (despite not particularly expecting this to change per-frame).

tmux/demos/medical/old/Kitware-M3.yml
josephvanpeltkw: @Purg, M3 was the next medical task that was targeted for improvement.

Purg (approved):

PR #475, opened 2024-11-20
add a detection of bagged source timestamps to fix latency calculation

josephvanpeltkw:
modify nodes to change the source stamp if it is coming from a bag


josephvanpeltkw:
Abandoning this method and switching to another method instead

PR #483, opened 2024-12-20
belief file csv output

josephvanpeltkw:
add new script to output the belief file csv

### DIFF
--- a/ros/angel_utils/scripts/bag_extractor.py
+++ b/ros/angel_utils/scripts/bag_extractor.py
@@ -130,7 +130,7 @@ class BagConverter(Node):
             .string_value
         )
         self.pv_image_frame_id = (
-            self.declare_parameter("pv_image_frame_id", "PVFramesBGR")
+            self.declare_parameter("pv_image_frame_id", "/kitware/pv_image_detections_2d")#"PVFramesBGR")
             .get_parameter_value()
             .string_value
         )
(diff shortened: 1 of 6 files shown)


ros/angel_utils/scripts/bag_extractor.py
```diff
@@ -130,7 +130,7 @@ def __init__(self):
             .string_value
         )
         self.pv_image_frame_id = (
-            self.declare_parameter("pv_image_frame_id", "PVFramesBGR")
+            self.declare_parameter("pv_image_frame_id", "/kitware/pv_image_detections_2d")#"PVFramesBGR")
```
Purg, line 133: I would have expected this to just need to be `"pv_image_detections_2d"`, and the namespace under which the bag play happens does appropriate namespace prepending to messages emitted?

ros/angel_utils/scripts/bag_extractor.py
```diff
@@ -570,6 +574,9 @@ def handle_image_msg(self, msg: Image) -> None:
             else:
                 rgb_image = bridge.imgmsg_to_cv2(msg, msg.encoding)
 
+            if "rgb" in msg.encoding:
+                rgb_image = cv2.cvtColor(rgb_image, cv2.COLOR_BGRA2RGBA)
```
Purg, line 578: Is L575 not outputting an RGB image already when the `msg.encoding` is "rgb"? It looks like  you're converting form "BGR" to "RGB" here, but I would have expected `rgb_image` to already be in RGB at this point (its just maybe missing the alpha channel?).

